PR #45, opened 2019-05-07
Introduces `common.py` to hold functions used across the codebase. `shell_execute` no longer merges stderr into stdout

lolski:
Introduces `common.py` to hold functions used across the codebase. This should reduce code duplication and make the codebase more maintainable

`shell_execute` no longer merges stderr into stdout. Instead it returns `stdout` and `stderr` in a separate variables. If exit code is non-zero, it will throw / raise a `RuntimeError`:
```
try:
    stdout, stderr = common.shell_execute(['bazel', 'test', '//...'])
except RuntimeError:
    ...
```

### DIFF
--- a/common.py
+++ b/common.py
@@ -21,12 +21,11 @@ def shell_execute(args, **kwargs):
             err = err.decode()
 
         if process.returncode == 0:
-            print('cmd = {}\nout = {}\nerr = {}'.format(args, out, err))
             return out, err
         else:
             raise RuntimeError('An error occurred.\n' +
                                '- Command = "'+ str(args) + '"\n' +
                                '- Exit code = "' + str(process.returncode) + '"\n' +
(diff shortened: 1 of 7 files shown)


common.py
```diff
@@ -0,0 +1,18 @@
...
+import subprocess as sp
+
+
+def shell_execute(*args):
+    p = sp.Popen(args, stdout=sp.PIPE, stderr=sp.PIPE)
+    p.wait()
+    out = p.stdout.read()
+    err = p.stderr.read()
+    if p.returncode == 0:
+        if type(out) == bytes:
```
vmax: @lolski `type(err)` could be `bytes` as well then?

lolski: under which condition would we get back bytes instead of string again, @vmax ?

vmax: @lolski when running script with Python 3

PR #48, opened 2019-05-10
Implement sync-version tools

### LINKED ISSUE #47: Create CI tool to Increment version number after a release
We want `//ci:sync-version` to bump the version of the LOCAL repository in which it is executed in. What that means is that if the current repository has `VERSION` file containing `X.Y.Z`, when `//ci:sync-version` is called, it will change the content of the file to be `X.Y.Z+1`. 

E.g. `VERSION` file content changes from `1.5.2` to `1.5.3`.

Now, we have the option of:
a) provide `/path/to/VERSION` as arguments to the `bazel run` command (e.g. `-- /path/to/VERSION`)
b) provide `/path/to/VERSION` as argument in the target declaration in the `BUILD` file
c) not provide `/path/to/VERSION` as it should always be expected (and thus hard-coded) to be at the root of every repository.

vmax:
Fix #47 

This is a tool that helps incrementing version in version file across @graknlabs repos. Currently, only incrementing **patch** (following [semantic versioning](https://semver.org) terminology) version is supported.

Sample usage:

```
py_binary(
    name= "sync_version",
    srcs = ["@graknlabs_build_tools//ci:sync-version.py"],
    data= ["//:VERSION"],
    main = "ci/sync-version.py",
    args = ["$(location //:VERSION)"]
)
```

```
ghost:grakn vmax$ bazel run //:sync_version
INFO: Analyzed target //:sync_version (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:sync_version up-to-date:
  bazel-bin/sync_version
INFO: Elapsed time: 0.200s, Critical Path: 0.01s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
[2019-05-10 12:56:50.316]: Read version 1.5.3 from version file ./VERSION
[2019-05-10 12:56:50.316]: Writing back version 1.5.4 into version file ./VERSION
```

```
diff --git a/VERSION b/VERSION
index 8af85beb5..94fe62c27 100644
--- a/VERSION
+++ b/VERSION
@@ -1 +1 @@
-1.5.3
+1.5.4
```

### DIFF
--- a/ci/sync-version.py
+++ b/ci/sync-version.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+
+#
+# GRAKN.AI - THE KNOWLEDGE GRAPH
+# Copyright (C) 2018 Grakn Labs Ltd
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+import sys
+import logging
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(
+    format='[%(asctime)s.%(msecs)03d]: %(message)s',
+    datefmt='%Y-%m-%d %H:%M:%S',
+)
+logger.level = logging.DEBUG
+
+if len(sys.argv) != 2:
+    raise Exception('Should pass version file as first argument')
+
+version = None
+version_fn = sys.argv[1]
+
+with open(version_fn) as version_file:
+    version = version_file.read().strip().split('.')
+
+if len(version) != 3:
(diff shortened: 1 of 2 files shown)


ci/sync-version.py
```diff
@@ -0,0 +1,50 @@
...
+    raise Exception('Should pass version file as first argument')
+
+version = None
+version_fn = sys.argv[1]
+
+with open(version_fn) as version_file:
+    version = version_file.read().strip().split('.')
+
+if len(version) != 3:
+    raise Exception('Version file should contain \'major.minor.patch\'')
```
lolski: How about `Version file should be in the following format: 'x.y.z'`? I think major, minor, patch isn't the correct term for us, semantically

ci/sync-version.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python
```
lolski: This should be Apache, right @vmax ?

vmax: @lolski `LICENSE` at the repo top says it's GNU AGPL v3, [ref](https://github.com/graknlabs/build-tools/blob/master/LICENSE)

ci/sync-version.py
vmax: @lolski fixed in https://github.com/graknlabs/build-tools/pull/48/commits/e62219b60e5a0c520985cf60cc85d649487d3439

ci/sync-version.py
lolski: You're right, let's go with this then.

lolski:
@haikalpribadi @vmax approved.

PR #51, opened 2019-05-21
Implement SingleJar wrapper and include it in custom Java toolchain

vmax:
This is the first step to fixing graknlabs/grakn#4851

`singlejar` wrapper (`@graknlabs_build_tools//bazel:singlejar`) detects when `singlejar` is called to build a source JAR and modifies command-line arguments so that source file is placed with respect to `package` statement in it.

For class `com.smth.Main` *incorrect* (from IntellIJ”s standpoint) structure (and this is something that is currently produced) would be:
```
ghost:b vmax$ jar tf bazel-bin/main-src.jar
META-INF/
META-INF/MANIFEST.MF
Main.java
```

Correct source JAR (that would actually be picked up) would look like this:
```
ghost:b vmax$ jar tf bazel-bin/main-src.jar
META-INF/
META-INF/MANIFEST.MF
com/
com/lib/
com/lib/Main.java
```

### DIFF
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -16,6 +16,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+load("@bazel_tools//tools/jdk:default_java_toolchain.bzl", "default_java_toolchain")
+
+
 java_binary(
     name = "bazel-deps",
     main_class = "com.github.johnynek.bazel_deps.ParseProject",
@@ -30,4 +33,28 @@ sh_binary(
         "@unused_deps//file",
         "@unused_deps_osx//file",
     ],
+)
+
+# Toolchain that redefines singlejar to a version
+# that produces IntelliJ-compatible source JARs
+default_java_toolchain(
+    name = "java-toolchain",
+    source_version = "8",
+    target_version = "8",
+    singlejar = select({
+        "@bazel_tools//src/conditions:remote": ["@bazel_tools//src/tools/singlejar:singlejar"],
+        "//conditions:default": [":singlejar_deploy.jar"]
+    }),
+    visibility = ["//visibility:public"],
+)
+
+# Wrapper that remaps .java files to have correct
+# placement inside the source JAR
+java_binary(
+    name = "singlejar",
+    srcs = ["SingleJar.java"],
+    resources = [
+        "@bazel_tools//tools/jdk:singlejar/singlejar"
+    ],
+    main_class = "grakn.buildtools.SingleJar"
 )
--- a/bazel/SingleJar.java
+++ b/bazel/SingleJar.java
@@ -1,0 +1,100 @@
+package grakn.buildtools;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class SingleJar {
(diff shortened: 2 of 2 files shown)


bazel/BUILD
```diff
@@ -30,4 +33,28 @@ sh_binary(
...
+)
+
+# Toolchain that redefines singlejar to a version
+# that produces IntelliJ-compatible source JARs
+default_java_toolchain(
+    name = "java-toolchain",
+    source_version = "8",
+    target_version = "8",
+    singlejar = select({
+        "@bazel_tools//src/conditions:remote": ["@bazel_tools//src/tools/singlejar:singlejar"],
```
lolski: I guess we will need to update `@bazel_tools//src/conditions:remote` with our own singlejar as well?

bazel/SingleJar.java
```diff
@@ -0,0 +1,100 @@
...
+                        args[i] = String.format("%s:%s", originalFn, resultingPath);
+                    }
+                }
+            }
+        }
+    }
+
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        for (String arg : args) {
```
lolski: what can `args` contain, @vmax ? can you give some examples?

lolski:
One more question, you mentioned that this PR is the first step to fixing the source JAR issue. What are the next steps which we still need to do? @vmax

bazel/BUILD
vmax: @lolski Not necessarily. Currently we only indent to resolve the issue of "Go to source" not working in IntelliJ — only local machines are affected.

lolski: @vmax is it as easy as this
```
"@bazel_tools//src/conditions:remote": [":singlejar_deploy.jar"]
```

or will it require more effort?

Ideally, we keep an identical implementation to avoid "bugs that only happen in local but not in RBE" and vice versa

bazel/SingleJar.java
vmax: @lolski `args` are passed to `singlejar`. The part we are interested in is `--resources`, after which "A.java:B.java" statements go — which map source files on filesystem to their location inside JAR

vmax:
@lolski the part would finally fix it is *using* that toolchain so the issue is resolved. for that, we'll need to merge this PR & update `.bazelrc` in `@graknlabs_grakn_core`

bazel/SingleJar.java
lolski: ok. is it in `source:destination` format?

ie., if `A.java` contains a class `grakn.core.A`, then are we changing the invocation from `A.java:A.java` to `A.java:grakn/core/A.java`?

vmax: @lolski your understanding is correct here

bazel/BUILD
vmax: @lolski actually, separate `singlejar` implementations are already used in stock config

```
vmax@ghost build-tools java-toolchain-with-modified-singlejar bazel query --output=build @bazel_tools//tools/jdk:singlejar                         
DEBUG: Rule 'io_bazel_rules_python' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1545256788 -0500"
# /private/var/tmp/_bazel_vmax/6866c2e533a140c6a78f612a435b4e8b/external/bazel_tools/tools/jdk/BUILD:145:1
filegroup(
  name = "singlejar",
  srcs = select({"@bazel_tools//src/conditions:remote": ["@bazel_tools//src/tools/singlejar:singlejar"], "//conditions:default": ["@bazel_tools//tools/jdk:singlejar/singlejar"]}),
)
```

I don't know the *exact* reason for this, but RBE will continue to use the same implementation as previously; we only replace locally-used `singlejar` — which would be relatively safe

lolski: ok, let's keep this in mind if we do encounter a weird JAR bug happening either only locally or in RBE.

PR #56, opened 2019-05-30
Use same singlejar binary on all platforms

vmax:
In order to support building Grakn on all platforms on both `v0.25.2` and `v0.26.0` of `bazel`, same `singlejar` should be used. This PR brings support for `.exe` versions of `singlejar` for Windows into our custom `singlejar` and unifies the source jar building process on all platforms.


lolski:
looks good

PR #78, opened 2019-07-22
Make sure builds executed locally don't fail

### LINKED ISSUE #77: bazel build //... should be runnable without RBE in order to support PRs from external forks
## Description

bazel build //... should be runnable without RBE in order to support PRs from external forks. Right now it would fail: https://circleci.com/gh/graknlabs/grakn/21383

vmax:
Fixes #77 
The logic is as follows:
- if build is *not* an RBE build:
  - if platform is Linux:
    - `apt-get install rpm`
  - if platform is Darwin:
    - `brew install rpm`
  - else
    - fail

### DIFF
--- a/ci/install-bazel-rbe.sh
+++ b/ci/install-bazel-rbe.sh
@@ -15,6 +15,8 @@ function install_dependencies() {
     fi
 }
 
+install_dependencies
+
 echo "Installing RBE credential..."
 if [[ -n "$BAZEL_RBE_CREDENTIAL" ]]; then
     BAZEL_RBE_CREDENTIAL_LOCATION=~/.config/gcloud/application_default_credentials.json
@@ -24,5 +26,4 @@ if [[ -n "$BAZEL_RBE_CREDENTIAL" ]]; then
     echo "The RBE credential has been installed!"
 else
     echo "No RBE credential found. Bazel will be executed locally without RBE support."
-    install_dependencies
 fi
--- a/ci/install-bazel-rbe.sh
+++ b/ci/install-bazel-rbe.sh
@@ -1,6 +1,20 @@
 #!/bin/bash
 
 set -e
+
+function install_rpmbuild() {
+    echo "Installing rpmbuild..."
+    if [[ "$OSTYPE" == "linux-gnu" ]]; then
+        sudo apt-get update
+        sudo apt-get install rpm
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+        brew install rpm
+    else
+        echo "Your platform does not have rpmbuild executable for it. Make sure you are using Linux/macOS".
+        exit 1
+    fi
+}
+
 echo "Installing RBE credential..."
 if [[ -n "$BAZEL_RBE_CREDENTIAL" ]]; then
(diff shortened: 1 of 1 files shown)


lolski:
are dependencies like these something that "Bazel toolchain" should solve for us?

vmax:
@lolski I don't think 'bazel toolchain' is responsible for _obtaining_ your tools, only for _locating_ them when they are already present

ci/install-bazel-rbe.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/bash
 
 set -e
+
+function install_rpmbuild() {
```
lolski: Let's call the function `install_dependencies` instead. Because it reads better from a high level

ci/install-bazel-rbe.sh
```diff
@@ -10,4 +24,5 @@ if [[ -n "$BAZEL_RBE_CREDENTIAL" ]]; then
     echo "The RBE credential has been installed!"
 else
     echo "No RBE credential found. Bazel will be executed locally without RBE support."
+    install_rpmbuild
```
lolski: Let's put the code outside of the if else, so that it is always installed regardless of whether you've RBE.

Is there a better file to host the code? I think `install-bazel-rbe.sh` should only be for installing RBE.

ci/install-bazel-rbe.sh
vmax: @lolski fixed in https://github.com/graknlabs/build-tools/pull/78/commits/1c3dc1b28ed7a6c165806bc4d24e1dc493c6ecfa

ci/install-bazel-rbe.sh
vmax: @lolski 
>it is always installed regardless of whether you've RBE.

This is redundant — RBE builds do not require `rpmbuild` installed and we'll lose ~15 seconds per each build

>Is there a better file to host the code? I think install-bazel-rbe.sh should only be for installing RBE.

`rpmbuild` is always needed, regardless of platform. In that way, I feel installing `rpmbuild` _only_ in `else` branch, when RBE is not present, is an implementation detail. 

A bigger idea for refactor might be merging install *all* needed deps for build (`bazel`, `rbe credential`, `rpmbuild`) in a script named `install-build-tools`

lolski: > This is redundant — RBE builds do not require rpmbuild installed and we'll lose ~15 seconds per each build

True, however imagine a user running Bazel command without RBE, in a machine which has RBE credential which will cause it to fail. That would be quite a confusing error. I think it's better to be slower for 15 seconds rather than be confusing.

> A bigger idea for refactor might be merging install all needed deps for build (bazel, rbe credential, rpmbuild) in a script named install-build-tools

This sounds like a good idea in the future. Can you create a Github issue for that?

vmax: @lolski addressed in https://github.com/graknlabs/build-tools/pull/78/commits/e80d9ee9c06241a881ac24a8426c8f46a8e24032

vmax: @lolski issue https://github.com/graknlabs/build-tools/issues/79 created

PR #89, opened 2019-09-10
Install rpmbuild conditionally

vmax:
To increase build time and not install `rpmbuild` which is not needed is build is executed remotely


haikalpribadi:
We'll be saving at least half the time of all our CI jobs @lolski !!! :) Thanks @vmax !

PR #95, opened 2019-09-23
Implement //ci:release-validate-deps

lolski:
The `//ci:release-validate-deps` can be used from the CI job to ensure that we're depending on a release instead of a snapshot dependency.

For example, if we are to release `graknlabs_grakn_core` which depends on a snapshot version of `graknlabs_common` will fail with the following error message:

```
bazel run @graknlabs_build_tools//ci:release-validate-deps graknlabs_graql graknlabs_common

RuntimeError: This commit is not releasable because there are one or more snapshot 
dependencies: ['graknlabs_common']. Check that every dependencies listed in 
dependencies/graknlabs/dependencies.bzl are all release dependencies 
(ie., depends on a tag instead of a commit).
```

If every dependency is a release dependency, it will succeed:
```
bazel run @graknlabs_build_tools//ci:release-validate-deps graknlabs_graql graknlabs_common

This repository is releasable because the dependencies
are all release dependencies: ['graknlabs_common', 'graknlabs_graql']
```

### DIFF
--- a/ci/release-validate-deps.py
+++ b/ci/release-validate-deps.py
@@ -1,0 +1,27 @@
+import os
+import sys
+
+if __name__ == '__main__':
+    dependencies_bzl = os.path.join('dependencies', 'graknlabs', 'dependencies.bzl')
+    snapshot_dependencies = []
+    with open(os.path.join(os.getenv("BUILD_WORKSPACE_DIRECTORY"), dependencies_bzl)) as dependencies_bzl_:
+        dependencies_bzl__ = dependencies_bzl_.read().splitlines()
+        for arg in sys.argv[1:]:
+            dependency = filter(lambda line: line.endswith('@' + arg), dependencies_bzl__)
+            if len(dependency) == 1:
+                depend_by_tag = dependency[0].strip().startswith('tag')
+                if not depend_by_tag:
+                    snapshot_dependencies.append(arg)
+            else:
+                raise RuntimeError('Invalid sync-marker found for {}. '
+                                   'There is supposed to be exactly one sync-marker '
+                                   'but {} occurrences found.'.format(arg, len(dependency)))
+    if len(snapshot_dependencies) == 0:
+        print('This repository is releasable because the dependencies are '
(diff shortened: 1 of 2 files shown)


ci/release-validate-deps.py
```diff
@@ -0,0 +1,27 @@
...
+import sys
+
+if __name__ == '__main__':
+    dependencies_bzl = os.path.join('dependencies', 'graknlabs', 'dependencies.bzl')
+    snapshot_dependencies = []
+    with open(os.path.join(os.getenv("BUILD_WORKSPACE_DIRECTORY"), dependencies_bzl)) as dependencies_bzl_:
+        dependencies_bzl__ = dependencies_bzl_.read().splitlines()
+        for arg in sys.argv[1:]:
+            dependency = filter(lambda line: line.endswith('@' + arg), dependencies_bzl__)
+            if len(dependency) == 1:
```
vmax: @lolski running this with Python 3 would cause `TypeError: object of type 'filter' has no len()` — pass it to `list()` first

ci/release-validate-deps.py
```diff
@@ -0,0 +1,27 @@
+import os
+import sys
+
+if __name__ == '__main__':
+    dependencies_bzl = os.path.join('dependencies', 'graknlabs', 'dependencies.bzl')
+    snapshot_dependencies = []
+    with open(os.path.join(os.getenv("BUILD_WORKSPACE_DIRECTORY"), dependencies_bzl)) as dependencies_bzl_:
```
vmax: @lolski better naming IMO would be `dependencies_bzl_fn` (filename) /`dependencies_bzl` (file object)

lolski: Resolved in ad7e6c531b89c5711e31d4ab3e64b650463e56ad

ci/release-validate-deps.py
lolski: Resolved in ad7e6c531b89c5711e31d4ab3e64b650463e56ad

PR #109, opened 2019-12-19
use JavaInfo instead of the now obsolete java attribute

vincent-cognite:
The assemble_maven rule seems to be broken with bazel > 1.0.0
fails with the following error:
`object of type 'JavaSkylarkApiProvider' has no field 'outputs'`
This PR aims at fixing that issue.

### DIFF
--- a/distribution/maven/rules.bzl
+++ b/distribution/maven/rules.bzl
@@ -269,8 +269,8 @@ def _assemble_maven_impl(ctx):
 
     # there is also .source_jar which produces '.srcjar'
     if hasattr(target, "java"):
-        jar = target.java.outputs.jars[0].class_jar
-        srcjar = target.java.outputs.jars[0].source_jar
+        jar = target[JavaInfo].outputs.jars[0].class_jar
+        srcjar = target[JavaInfo].outputs.jars[0].source_jar
     elif hasattr(target, "files"):
         jar = target.files.to_list()[0]
     else:


haikalpribadi:
Thanks for the PR, @vincent-cognite! But just to check, are you using this repository to build non-@graknlabs repositories? We definitely do not mind (as it is open source), but do keep in mind that this repository is for graknlabs-specific projects, and is not meant to generalise to other organisations.

vincent-cognite:
Yes, we (just) started to use graknlabs/build-tools,bazel-distribution for deploying some bazel artifacts as maven artifacts in my organization. We are new to bazel, so a lot of (steep) learning!
graknlabs/bazel-distribution is mentioned here https://awesomebazel.com/ in the tooling section.

haikalpribadi:
> We are new to bazel, so a lot of (steep) learning!
We absolutely know what you mean, @vincent-cognite! It was also a steep learning curve for us (but it's worth it, keep going!)

> graknlabs/bazel-distribution is mentioned here https://awesomebazel.com/ in the tooling section.
Correct. And https://github.com/graknlabs/bazel-distribution is definitely built generically for public usage. So you should use it, and we will make sure to only make changes that maintain general usage for public.

However, https://github.com/graknlabs/build-tools is a repository of graknlabs-specific Bazel commands. You are free to use it, and you can definitely take inspiration from it. However, we will not be maintaining this codebase for public usage, and we will definitely be making changes to it that are specific to @graknlabs, and this may break your build system. I'll try to make this more clear in the repository description.

vincent-cognite:
I see. Thanks for letting me know @haikalpribadi !

haikalpribadi:
You're welcome @vincent-cognite !

vmax:
@vincent-cognite thank you for the PR! You're listed as an author here: https://github.com/graknlabs/bazel-distribution/commit/21df5bdaf732300fad48519ef9ba4eca5c20a7f8

Replaced by #115

PR #110, opened 2020-01-10
Update rules_proto

vmax:
In order to use upgraded `grpc`, we need to switch to our fork of `rules_proto` which has updated link to `io_grpc_grpc_java`


vmax:
Replaced by #112

PR #160, opened 2020-07-13
Change extract artifact to executable rule

adammitchelldev:
We re-discovered a way of making the extract script a bazel executable rule.


lolski:
The extraction is done using `tar` and therefore it won't be runnable in Windows. It means that we won't be able to run this rule on `test-assemble-windows-zip` for example. What do you think @haikalpribadi ?

PR #165, opened 2020-08-04
Fix pyenv to be cross-compatible between CircleCI and Grabl

vmax:
* `-s` option ignores potentially-installed version whch helps with reruns
* setting Python2's version as one that's installed in system helps us be cross-compatible with both CircleCI and Grabl

### DIFF
--- a/tool/bazelinstall/rbe.sh
+++ b/tool/bazelinstall/rbe.sh
@@ -43,8 +43,8 @@ if [[ -n "$BAZEL_BUILDBUDDY_CERT" && -n "$BAZEL_BUILDBUDDY_KEY" ]]; then
     echo "The RBE credential has been installed!"
     echo "Configuring Python..."
     # setting the exact version of Python 3 and Python 2, respectively
-    pyenv install 3.6.10
-    pyenv global 3.6.10 2.7.12
+    pyenv install -s 3.6.10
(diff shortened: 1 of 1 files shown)


tool/bazelinstall/rbe.sh
```diff
@@ -43,8 +43,8 @@ if [[ -n "$BAZEL_BUILDBUDDY_CERT" && -n "$BAZEL_BUILDBUDDY_KEY" ]]; then
     echo "The RBE credential has been installed!"
     echo "Configuring Python..."
     # setting the exact version of Python 3 and Python 2, respectively
-    pyenv install 3.6.10
-    pyenv global 3.6.10 2.7.12
+    pyenv install -s 3.6.10
```
lriuui0x0, line 46: Can we instead do:

```
pyenv install -s 3.6.10
pyenv install -s 2.7.12
pyenv global 3.6.10 2.7.12
```

vmax: @lriuui0x0 we could, but `python` (Python 2) is installed system-wide and we don't really care which one it is - we're going to move off depending on Python 2 soon anyway

lriuui0x0: OK cool @vmax

PR #183, opened 2020-08-20
Update grakn-bin and deps for Grakn 2.0, add netrc auth

lolski:
Update grakn-bin and deps for Grakn 2.0, add netrc auth  for artifact repository, allowing us to push/read private artifact repositories.


distribution/deps.bzl
```diff
@@ -17,10 +17,17 @@
...
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 def deps():
     git_repository(
         name = "graknlabs_bazel_distribution",
         remote = "https://github.com/graknlabs/bazel-distribution",
-        commit = "d62ce06aacb1f185b1be737b7e6db8168568aa5e" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_bazel_distribution
+        commit = "17aabcf469e95e572e8c40b0b52f2dfd6b55bbef" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_bazel_distribution
     )
+    http_archive(
```
lolski, line 28: what do we skylib for? do you remember @adammitchelldev @flyingsilverfin ?

tool/checkstyle/templates/checkstyle.xml
```diff
@@ -100,9 +100,9 @@
             maven.willGetAngry();
         ```
         -->
-        <module name="NeedBraces">
-            <property name="allowSingleLineStatement" value="true"/>
-        </module>
+        <!--<module name="NeedBraces">-->
```
lolski, line 103: why are we commenting this out @adammitchelldev ?

flyingsilverfin (approved):

binary/grakn
```diff
@@ -23,7 +23,7 @@ GRAKN_HOME=$(cd "$(dirname "${path}")" && pwd -P)
 GRAKN_CONFIG="server/conf/grakn.properties"
 
 CONSOLE_JAR_FILENAME=(${GRAKN_HOME}/console/services/lib/io-grakn-console-grakn-console*.jar)
-SERVER_JAR_FILENAME=(${GRAKN_HOME}/server/services/lib/io-grakn-core-grakn-server*.jar)
+SERVER_JAR_FILENAME=(${GRAKN_HOME}/server/lib/grakn-core-grakn-core-server*.jar)
```
haikalpribadi, line 26: This change should never have happened, @lolski . You only needed to make this change because of the mistake we made here: https://github.com/graknlabs/grakn-2.0/pull/7#pullrequestreview-478152772

But this should be a code-smell to pay attention to next time, @lolski; if we have to rename something that used to work all this while, perhaps something just broke to cause this change? That would have helped us notice that we named the artifacts incorrectly in Grakn Core 2.0.

PR #193, opened 2020-09-01
Update junit from 4.12 to 4.13

jmsfltchr:


jmsfltchr:
I was trying this upgrade in order to use `@BeforeParams` for parameterised testing of `graknlabs/simulation`, but this is no longer needed as I found a different architecture for the solution.

PR #194, opened 2020-09-02
Allow, and force, license header test to run on non-java files

alexjpwalker:
## Problem to Solve

Some of the license headers were wrong in client-java, but the build still passed because we only verify the license in Java source files.

## Proposed Solution

- Remove the restriction that only *.java files have their license headers checked by checkstyle
- Allow any file visible to a BUILD file (including the BUILD file itself, which is the primary use case here) to be added to a new array property of `checkstyle_test` named `files`
- Enforce that every `checkstyle_test` must specify which license is applicable - don't just assume `agpl`
- Enforce that every BUILD file and .bzl file must be the target of a `checkstyle_test`

## Usage notes

Sample usage as requested by @haikalpribadi 
```
checkstyle_test(
    name = "checkstyle",
    targets = [":client-java"],
    files = ["BUILD"],
    license_type = "apache",
)
```
The expectation is that these new properties (`files` and the now-mandatory `license_type`) will be added to every BUILD file that requires a license.


haikalpribadi (approved):
The UX looks good. Assigning @flyingsilverfin to review the code.

tool/checkstyle/rules.bzl
```diff
@@ -22,18 +22,18 @@ JavaSourceFiles = provider(
...
+def collect_java_sources_impl(target, ctx):
+    javaSources = []
     if hasattr(ctx.rule.attr, 'srcs'):
         for src in ctx.rule.attr.srcs:
             for f in src.files.to_list():
                 if f.extension == 'java':
-                    files.append(f)
-    return [JavaSourceFiles(files = files)]
+                    javaSources.append(f)
+    return [JavaSourceFiles(files = javaSources)]
```
flyingsilverfin, line 32: should rename "java" everywhere to just "source"

tool/checkstyle/rules.bzl
```diff
@@ -76,17 +76,21 @@ def _checkstyle_test_impl(ctx):
       args += " -p %s" % properties.path
       inputs.append(properties)
 
-    all_java_files = []
+    files = []
     for target in ctx.attr.targets + [ctx.attr.target]:
         if target:
-            all_java_files.extend(target[JavaSourceFiles].files)
+            files.extend(target[JavaSourceFiles].files)
+    print(files)
```
flyingsilverfin, line 83: stray print

tool/checkstyle/rules.bzl
```diff
@@ -76,17 +76,21 @@ def _checkstyle_test_impl(ctx):
...
+    for target in ctx.attr.files:
+        if target:
```
flyingsilverfin, line 85: why do we need this `if` check?

tool/checkstyle/rules.bzl
```diff
@@ -131,11 +135,15 @@ checkstyle_test = rule(
         ),
         "target": attr.label(
             doc = "The java_library target to check sources on",
```
flyingsilverfin, line 137: same, no more "java" here or in the next attribute

flyingsilverfin (changes requested):

tool/checkstyle/rules.bzl
alexjpwalker: oops

tool/checkstyle/rules.bzl
alexjpwalker: fixed

flyingsilverfin (approved):

PR #203, opened 2020-09-29
Add RocksDB JNI dev build tool

alexjpwalker:

To automate the build process for RocksDB JNI (dev version) so it is less error-prone and faster

Add 2 Bazel targets:

- `assemble_maven` - for building RocksDB JNI-dev into a Maven artifact
- `deploy_maven` - for deploying it to our Maven repository

And add a README file explaining what they do and how to execute them.

## Additional information

In short, the process for deploying RocksDB JNI-dev is:

1. `bazel build //library/rocksdbjni:assemble-maven`
2. `bazel run //library/rocksdbjni:deploy-maven -- release`


library/maven/artifacts.bzl
```diff
@@ -213,8 +214,5 @@ artifacts = {
     "org.slf4j:slf4j-simple": "1.7.20",
     "org.yaml:snakeyaml": "1.25",
     "org.zeromq:jeromq": "0.5.2",
-    "org.zeroturnaround:zt-exec": {
-        "exclude": ["commons-io:commons-io"],
-        "version": "1.10",
-    },
+    "org.zeroturnaround:zt-exec": "1.10",
```
alexjpwalker, line 217: This is needed in order for RocksDbBuilder.kt to execute successfully

distribution/deps.bzl
```diff
@@ -21,5 +21,5 @@ def graknlabs_bazel_distribution():
     git_repository(
         name = "graknlabs_bazel_distribution",
         remote = "https://github.com/graknlabs/bazel-distribution",
-        commit = "b6ae4297ab66484107f9469d88a4e5d047b36c53" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_bazel_distribution
+        commit = "d52066841775b84c5858ca35b8ac259a20a9eb4d" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_bazel_distribution
```
alexjpwalker, line 24: This allows `assemble_maven`, `deploy_maven` to take in a `java_import`

library/rocksdbjni/BUILD
```diff
@@ -0,0 +1,59 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_binary")
+load("@graknlabs_bazel_distribution//maven:rules.bzl", "assemble_maven", "deploy_maven")
+load("@graknlabs_dependencies//distribution:deployment.bzl", "deployment")
+
+kt_jvm_binary(
```
lolski, line 24: let's call the target `buidler` to be consistent?

`bazel build //library/rocksdbjni:builder` sounds nicer than `bazel build //library/rocksdbjni:build-rocksdb`

library/rocksdbjni/BUILD
```diff
@@ -0,0 +1,59 @@
...
+    name = "build-rocksdb",
+    main_class = "library.rocksdbjni.RocksDbBuilderKt",
+    srcs = ["RocksDbBuilder.kt"],
+    deps = [
+        # Maven dependencies
+        "@maven//:org_zeroturnaround_zt_exec"
+    ]
+)
+
+genrule(
```
lolski, line 34: let's call the target `builder` to be consistent?

`bazel build //library/rocksdbjni:builder` sounds nicer than `bazel build //library/rocksdbjni:build-rocksdb`

library/rocksdbjni/BUILD
```diff
@@ -0,0 +1,59 @@
...
+    name = "rocksdbjni-jar",
+    outs = ["rocksdbjni-osx.jar"],
+    srcs = ["VERSION"],
+    cmd = "$(execpath :build-rocksdb) && mv rocksdbjni-osx.jar \"$@\"",
+    tools = ["@bazel_tools//tools/jdk:current_host_java_runtime", ":build-rocksdb"],
+)
+
+java_import(
```
lolski, line 42: Let's call this one, `jar`, `bazel build //library/rocksdbjni:jar` perhaps?

library/rocksdbjni/RocksDbBuilder.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import org.zeroturnaround.exec.ProcessExecutor
+import org.zeroturnaround.exec.ProcessResult
+import java.io.File
+import java.nio.file.Paths
+
+fun main() {
+    val baseDir = Paths.get(".")
+    val version = Paths.get("library").resolve("rocksdbjni").resolve("VERSION").toFile().useLines { it.firstOrNull() }
+    val javaHome = "/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home"
```
lolski, line 11: There's no way to generify it huh? Do you have any idea if we can supply the argument from outside for this one, @vmax ?

alexjpwalker: Arguments can't be supplied to a `genrule`. I raised an [issue](https://github.com/graknlabs/dependencies/issues/204) for someone to take a look at this in the future.

library/rocksdbjni/BUILD
alexjpwalker: Yeah this is the trouble, `build-rocksdb` builds the JAR, and this genrule runs build-rocksdb, hence building the JAR.
Perhaps we should call the Kotlin target `build-script` and the genrule `builder`?

library/rocksdbjni/BUILD
alexjpwalker: See below

library/rocksdbjni/BUILD
lolski: Looks much better!

lolski (approved):

library/rocksdbjni/BUILD
```diff
@@ -0,0 +1,64 @@
...
+        "rocksdbjni-sources.jar",
+    ],
+    srcs = ["VERSION"],
+    cmd = "$(execpath :build-script) && read -a outs <<< '$(OUTS)' && mv rocksdbjni-osx.jar $${outs[0]} && mv rocksdbjni-sources.jar $${outs[1]}",
+    tools = ["@bazel_tools//tools/jdk:current_host_java_runtime", ":build-script"],
+    tags = ["manual"],
+)
+
+java_import(
+    name = "rocksdbjni-java-lib",
```
lolski, line 47: Should we rename this to `lib`?

`bazel build //library/rocksdbjni:lib` feels nicer than ``bazel build //library/rocksdbjni:rocksdbnji-java-lib`

library/maven/artifacts.bzl
```diff
@@ -204,7 +204,8 @@ artifacts = {
     "org.mockito:mockito-core": "2.6.4",
     "org.openjdk.jmh:jmh-core": "1.23",
     "org.openjdk.jmh:jmh-generator-annprocess": "1.23",
-    "org.rocksdb:rocksdbjni": "6.8.1",
+    "org.rocksdb:rocksdbjni": "6.11.4",
+    "org.rocksdb:rocksdbjni-dev": "6.11.4",
```
lolski, line 208: Given that we're outputing `org.rocksdb:rocksdbjni-dev`, should we rename the whole package to `rocksdbjni-dev`?

lolski (commented):
Let's not include these targets when a user does `bazel build //...`. I think it's done by adding `tags=["manual"]` to the targets?

Because otherwise, it would fail depending on where they have the Java path (we can already see that it's failing on CI). Let's also include that fact in the issue that you've created, so that the next person can fix it.

library/rocksdbjni/BUILD
```diff
@@ -0,0 +1,64 @@
...
+
+genrule(
+    name = "builder",
+    outs = [
+        "rocksdbjni-osx.jar",
+        "rocksdbjni-sources.jar",
+    ],
+    srcs = ["VERSION"],
+    cmd = "$(execpath :build-script) && read -a outs <<< '$(OUTS)' && mv rocksdbjni-osx.jar $${outs[0]} && mv rocksdbjni-sources.jar $${outs[1]}",
+    tools = ["@bazel_tools//tools/jdk:current_host_java_runtime", ":build-script"],
```
lolski, line 42: Given that we're importing java here, can we perhaps use it in the Kotlin script? That way, we're using a Bazel that's guaranteed to be included by Bazel.

```
kt_jvm_library(
    name = "build-script",
    main_class = "library.rocksdbjni.RocksDbBuilderKt",
    srcs = ["RocksDbBuilder.kt"],
    deps = [
        "@bazel_tools//tools/jdk:current_host_java_runtime",
        # Maven dependencies
        "@maven//:org_zeroturnaround_zt_exec"
    ]
)

File: library/maven/artifacts.bzl (hunk 8 above)
Comment (alexjpwalker) [reply library/maven/artifacts.bzl:208]: `org.rocksdb` is the name chosen by the RocksDB team which fits with standard package naming guidelines (i.e. take a domain name and reverse it). I think I prefer `rocksdb` - note that the packages in Java code are also named `org.rocksdb`, and it is only intended that we import one JAR at a time.

File: library/rocksdbjni/BUILD (hunk 9 above)
Comment (alexjpwalker) [reply library/rocksdbjni/BUILD:42]: Sure, I'll try try moving it there

Comment (alexjpwalker) [reply library/rocksdbjni/BUILD:42]: It gives:
```
ERROR: /Users/aw/Desktop/workspace/dependencies/library/rocksdbjni/BUILD:24:14: in deps attribute of kt_jvm_binary rule //library/rocksdbjni:build-script: '@bazel_tools//tools/jdk:current_host_java_runtime' does not have mandatory providers: 'JavaInfo'
```

alexjpwalker:
> Let's not include these targets when a user does `bazel build //...`. I think it's done by adding `tags=["manual"]` to the targets?
> 
> Because otherwise, it would fail depending on where they have the Java path (we can already see that it's failing on CI). Let's also include that fact in the issue that you've created, so that the next person can fix it.

@lolski I initially thought the same but it didn't work - according to max, `tags=["manual"]` is only respected by `bazel test` (just like all those other useful parameters like `test_output=streamed`, haha......)

So I updated `automation.yml` to call `bazel build //... --deleted_packages=library/rocksdbjni` and now we have our green tick.

Not ideal but there is no other way. I tried `.bazelignore` but then it's not even possible to build rocksdb manually.

library/rocksdbjni/BUILD
alexjpwalker: I'll shorten the name to `java-lib`, giving us `bazel build //library/rocksdbjni:java-lib`.

PR #209, opened 2020-10-14
Allow short SHAs in snapshot artifacts

alexjpwalker:

To allow Grakn to run in Linux.

See https://github.com/graknlabs/bazel-distribution/pull/265.

Update bazel-distribution


alexjpwalker:
We decided not to do this as it would introduce an unacceptable paradigm shift

PR #213, opened 2020-10-26
Remove hard-coded Java path

lolski:

We have removed the hard-coded `JAVA_HOME` in `RocksDbBuilder.kt`. It is now retrieved dynamically with `/usr/libexec/java_home`.

This is something that wasn't possible before and is only made possible by making Java available within the sandbox. Specifically, we're talking about adding Java toolchain in the `tools` attribute (which apparently was already added in the [original PR](https://github.com/graknlabs/dependencies/pull/203) for a seemingly different reason):

```
genrule(
    name = "builder",
    ...
    tools = ["@bazel_tools//tools/jdk:current_host_java_runtime", ":build-script"],
)
```

### DIFF
--- a/library/rocksdbjni/RocksDbBuilder.kt
+++ b/library/rocksdbjni/RocksDbBuilder.kt
@@ -8,8 +8,9 @@ import java.nio.file.Paths
 fun main() {
     val baseDir = Paths.get(".")
     val version = Paths.get("library").resolve("rocksdbjni").resolve("VERSION").toFile().useLines { it.firstOrNull() }
-
-    val javaHome = shellScript("/usr/libexec/java_home", Paths.get(".").toFile(), ".")!!.outputUTF8().trim()
+    val javaHome = Paths.get(
+            shellScript("/usr/libexec/java_home", baseDir.toFile(), null)!!.outputUTF8().trim()
+    ).toFile()
 
     shellScript("git clone git@github.com:facebook/rocksdb.git", baseDir.toFile(), javaHome)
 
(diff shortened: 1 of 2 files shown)


library/rocksdbjni/RocksDbBuilder.kt
```diff
@@ -2,28 +2,28 @@ package library.rocksdbjni
...
 import org.zeroturnaround.exec.ProcessResult
-import java.io.File
+import java.nio.file.Path
 import java.nio.file.Paths
 
 fun main() {
     val baseDir = Paths.get(".")
     val version = Paths.get("library").resolve("rocksdbjni").resolve("VERSION").toFile().useLines { it.firstOrNull() }
-    val javaHome = "/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home"
+    val javaHome = Paths.get(shellScript("/usr/libexec/java_home", baseDir, null)!!.outputUTF8().trim())
```
alexjpwalker, line 11: What does this line do? Specifically the `!!.`

library/rocksdbjni/RocksDbBuilder.kt
```diff
@@ -35,10 +35,15 @@ fun main() {
     sourcesJar.copyTo(sourcesDestPath)
 }
 
-fun shellScript(cmd: String, baseDir: File, javaHome: String): ProcessResult? {
-    println(cmd)
-    return ProcessExecutor(cmd.split(" "))
-            .redirectOutput(System.out).redirectError(System.err)
-            .environment("JAVA_HOME", javaHome)
-            .directory(baseDir).execute()
+fun shellScript(script: String, baseDir: Path, javaHome: Path?): ProcessResult? {
```
alexjpwalker, line 38: Yeah, this is definitely cleaner

alexjpwalker (approved):

library/rocksdbjni/RocksDbBuilder.kt
lolski: That's how you invoke a method on a variable that may be `null`. I also didn't like it because it looks terse.

Let me try to find if there's a clearer way to do it (maybe by preceding the call with some kind of `assertNotNull` or something like that)

lolski: Actually we can clean it!! The original function returns a `Nullable` but actually it doesn't need to. Fixed in https://github.com/graknlabs/dependencies/pull/213/commits/7f55e0882571e8475a475660447200569483cba6

library/rocksdbjni/RocksDbBuilder.kt
```diff
@@ -4,11 +4,13 @@ import org.zeroturnaround.exec.ProcessExecutor
 import org.zeroturnaround.exec.ProcessResult
 import java.nio.file.Path
 import java.nio.file.Paths
+import kotlin.test.assertNotNull
```
alexjpwalker, line 7: Are we using this?

lolski: Good spot, removed in https://github.com/graknlabs/dependencies/pull/213/commits/867af6f633a8987ab61a945f505b19bac24dd58b

PR #216, opened 2020-11-09
Add a macro for loading Google OR Tools in a platform independent manner

lolski:

Google OR-Tools is not platform-independent - it has separate library files for Mac, Linux and Windows. Therefore, it would be ideal if we can allow the user to load the library from the respective repository without having to deal with making sure it works for the right platform.

The solution we came up with is by adding a Bazel macro for loading the library depending on the platform. For example, if you invoke Bazel in a Mac machine, it will correctly load the library meant for Mac. The same applies to the other operating systems.


library/maven/artifacts.bzl
```diff
@@ -227,4 +227,4 @@ artifacts = {
     "org.yaml:snakeyaml": "1.25",
     "org.zeromq:jeromq": "0.5.2",
     "org.zeroturnaround:zt-exec": "1.10",
-}
+}
```
alexjpwalker, line 230: missing newline
```suggestion
}

```

library/ortools/BUILD
```diff
@@ -1,47 +1,20 @@
...
+        "@platforms//os:linux",
+    ],
 )
+
+config_setting(
+    name = "is_windows",
+    constraint_values = [
+        "@platforms//os:windows",
+    ],
+)
```
alexjpwalker, line 20: missing newline
```suggestion
)

```

library/ortools/deploy/BUILD
```diff
@@ -0,0 +1,47 @@
...
+        "@google_or_tools_linux//:ortools-java-" + version + "-javadoc.jar",
+
+        "@google_or_tools_windows//:pom-runtime.xml",
+        "@google_or_tools_windows//:ortools-win32-x86-64-" + version + ".jar",
+        "@google_or_tools_windows//:ortools-win32-x86-64-" + version + "-sources.jar",
+        "@google_or_tools_windows//:pom-local.xml",
+        "@google_or_tools_windows//:ortools-java-" + version + ".jar",
+        "@google_or_tools_windows//:ortools-java-" + version + "-sources.jar",
+        "@google_or_tools_windows//:ortools-java-" + version + "-javadoc.jar",
+    ]
```
alexjpwalker, line 46: Feels like there's a fair bit of duplication here, not clear how to avoid it though

library/ortools/deploy/deps.bzl
```diff
@@ -0,0 +1,63 @@
...
+"""
+exports_files([
+    "ortools-win32-x86-64-{0}-sources.jar",
+    "ortools-win32-x86-64-{0}.jar",
+    "pom-runtime.xml",
+
+    "pom-local.xml",
+    "ortools-java-{0}.jar",
+    "ortools-java-{0}-sources.jar",
+    "ortools-java-{0}-javadoc.jar"
```
alexjpwalker, line 60: There's a fair bit of duplication here as well

alexjpwalker (approved):
It feels like there is room for improvement, perhaps making some of the rules more concise, but if it works and allows us to build Grakn Core then we should go ahead and merge this and perhaps brush it up later. Left a couple of suggested changes which are minor edits

library/maven/artifacts.bzl
lolski: Fixed

library/ortools/BUILD
lolski: Fixed

library/ortools/deploy/BUILD
lolski: What do you say about this approach: https://github.com/graknlabs/dependencies/pull/216/commits/08944c10fdf20b9841a698ac3f749c39b67a8b9a

library/ortools/deploy/deps.bzl
lolski: What do you say about this approach: https://github.com/graknlabs/dependencies/pull/216/commits/08944c10fdf20b9841a698ac3f749c39b67a8b9a

library/ortools/deploy/BUILD
lolski: with an additional renaming here: https://github.com/graknlabs/dependencies/pull/216/commits/fed7ef86c4bfed978c1f0bc7be337805f686f2c9

library/ortools/deploy/deps.bzl
lolski: with an additional renaming here: https://github.com/graknlabs/dependencies/pull/216/commits/fed7ef86c4bfed978c1f0bc7be337805f686f2c9

PR #220, opened 2020-11-13
Add tooling for importing platform-specific artifacts

alexjpwalker:

Since Grakn artifacts became platform-specific, we have difficulty running the client-java tests on our Mac machines. This PR addresses the issue by adding a macro that will expand to multiple platform-specific `artifact_file` rules.

Provided the Grakn Core artifact is not too large, it's a fair tradeoff to pull in more stuff even if it won't all be needed at runtime.

- Add `native_artifact_files` rule
- Move OS detection utils to `util.platform`


library/ortools/BUILD
```diff
@@ -1,20 +1 @@
-config_setting(
```
alexjpwalker, line 1: OS detection utilities were located in `library.ortools`. They are more generic than that, and should be relocated to `util.platform`.

distribution/artifact/rules.bzl
```diff
@@ -0,0 +1,10 @@
+load("@graknlabs_bazel_distribution//artifact:rules.bzl", "artifact_file")
+
+def native_artifact_files(name, artifact_name, **kwargs):
+    for platform in ["mac", "linux"]:
+        artifact_file(
+            name = name + "_" + platform,
+            # Can't use .format() because the result string will still have the unresolved parameter {version}
+            artifact_name = artifact_name.replace("{platform}", platform),
+            **kwargs,
+        )
```
alexjpwalker, line 10: New Bazel macro to pull in both the Mac and Linux artifacts into the WORKSPACE of a client

lolski (approved):

PR #223, opened 2020-11-25
Downgrade GRPC

lriuui0x0:

The upgraded GRPC version causes the deletion of the databases on linux grakn-2.0 hangs. We need to downgrade it.

Downgrade GRPC to the previous version 1.24.1


lolski:
@lriuui0x0 I'm going to merge it in order to unblock you on `client-java`. However, let's continue investigating on why exactly it is happening

lolski (approved):

PR #232, opened 2020-12-03
Fix dependency-analysis for good

vmax:

Currently, even successful executions of `dependency-analysis` fail the job. It happens because we expect the code to be 
`202` rather than `201`.

* Accept `201` as successful code
* Ease debugging by printing the output we get from `curl`


grabl (approved):

vmax:
@grabl thank you, virtual friend!

PR #237, opened 2020-12-17
Implement //tool/release:bump

vmax:

Currently, there are some post-release tasks that are done manually, namely:
- bump the version in `VERSION` file
- create a new GitHub milestone

Implement `//tool/release:bump` that would automate such tasks for us


tool/release/Bump.kt
```diff
@@ -0,0 +1,48 @@
...
+    println("Bumping the version to ${newVersion}")
+    Files.write(versionFile, newVersion.toByteArray())
+
+    println("Creating new milestone ${newVersion} in ${repoOwner}/${repoName}")
+    postJson("https://api.github.com/repos/${repoOwner}/${repoName}/milestones",
+            "Token ${githubToken}",
+            "application/vnd.github.v3+json",
+            """{"title": "${newVersion}"}""")
+
+}
```
lolski, line 48: can you add newline here @vmax ?

tool/release/Bump.kt
```diff
@@ -0,0 +1,48 @@
...
+fun postJson(url: String?, authorization: String, accept: String, content: String) {
+    NetHttpTransport()
+    .createRequestFactory()
+    .buildPostRequest(GenericUrl(url), ByteArrayContent.fromString("application/json", content))
+    .setHeaders(HttpHeaders().setAuthorization(authorization).setAccept(accept))
+    .execute()
+}
+
+fun main() {
+    val workspaceDirectory = System.getenv("BUILD_WORKSPACE_DIRECTORY")
```
lolski, line 20: these are good early assertions with informative error messages. I especially liked "Not running from within Bazel workspace". good stuff.

tool/release/Bump.kt
```diff
@@ -0,0 +1,48 @@
...
+    val repoOwner = System.getenv("GRABL_OWNER")
+            ?: throw RuntimeException("GRABL_OWNER environment variable is not set")
+    val repoName = System.getenv("GRABL_REPO")
+            ?: throw RuntimeException("GRABL_REPO environment variable is not set")
+
+    val versionFile = Paths.get(workspaceDirectory, "VERSION")
+    val content = String(Files.readAllBytes(versionFile)).trim()
+    val versionComponents = content.split(".").map { it -> Integer.parseInt(it)}.toTypedArray()
+
+    if (versionComponents.size != 3) {
```
lolski, line 33: our versioning is quite free now. can we update the validation to this regex: https://github.com/graknlabs/bazel-distribution/blob/master/maven/templates/deploy.py#L116

lolski: - if the version looks like `2.0.0`, bump it to `2.0.1`
- if it looks like `2.0.0-alpha-1` bump it to to `2.0.0-alpha-2`

tool/release/Bump.kt
vmax: @lolski addressed in https://github.com/graknlabs/dependencies/pull/237/commits/a91e4509ff8f42675ad580eec038602fb996b6f6

tool/release/Bump.kt
vmax: @lolski addressed in https://github.com/graknlabs/dependencies/pull/237/commits/b1039df806934c93deb04cee8be6f435f0927244

tool/release/Bump.kt
```diff
@@ -0,0 +1,66 @@
...
+
+fun postJson(url: String?, authorization: String, accept: String, content: String) {
+    NetHttpTransport()
+    .createRequestFactory()
+    .buildPostRequest(GenericUrl(url), ByteArrayContent.fromString("application/json", content))
+    .setHeaders(HttpHeaders().setAuthorization(authorization).setAccept(accept))
+    .execute()
+}
+
+fun increaseVersion(version: String): String {
```
lolski, line 19: can we call it `bumpVersion` (I think increase and bump has the same meaning in this context). We want to always reduce the number of terminology to the minimum required.

lolski: I think the algorithm can be simplified:

```
fun bumpVersion(version: String): String {
  current = version.split('.').getLast()
  current = current.split('-').getLast()
  try {
    next = parseInt(current) + 1
  }
  catch (e: NumberFormatException) {
    throw RuntimeException('unparseable ...')
  }
}

Comment (lolski) [reply tool/release/Bump.kt:19]: where's the case where you handle version like `2.0.0`?

Comment (vmax) [reply tool/release/Bump.kt]: @lolski addressed: https://github.com/graknlabs/dependencies/pull/237/commits/1a9419eb196abf2a85a6df22fa7261dbe7eccf78

Comment (vmax) [reply tool/release/Bump.kt]: @lolski it's handled in the very first line (as it's the most basic case):

```
        (Integer.parseInt(version) + 1).toString()
```

lolski (approved):

lolski (approved):

PR #239, opened 2020-12-20
Implement release_validate_nodejs_deps

vmax:

Currently, it's possible to release a NodeJS package which still depends on a snapshot version (say, from `repo.grakn.ai`). In order to prevent it, we need to run a test which goes over `package.json` and verifies that dependencies are properly versioned.

Implement `release_validate_nodejs_deps`; sample usage:

```
release_validate_nodejs_deps(
    name = "release-validate-nodejs-deps",
    package_json = "//:new-package.json",
    tagged_deps = ["grakn-protocol"]
)
```


tool/release/ValidateNodeJsDeps.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        val url = npmjsAddress.format(it, it, version)
+        try {
+            httpGet(url)
+        } catch (e: Exception) {
+            throw RuntimeException("cannot download version $version of $it (url $url); please ensure it has been released to npmjs.org")
+        }
+    }
+
+}
```
lolski, line 44: can you add a new line here? it's our convention which should be added to checkstyle at some point.

PR #251, opened 2021-01-21
Fix console warnings caused by out of date dependency

lriuui0x0:

Fix `WARNING: An illegal reflective access operation has occurred` warning caused by an out of date version of protobuf.

- Update protobuf to latest.


vmax (approved):
@lriuui0x0 approving it in hopes nothing would break since it's the backbone of all repos :slightly_smiling_face:

PR #260, opened 2021-02-17
Implement artifact_repackage

vmax:

Implement a macro that would assist in building distributions that include Grakn Console.

Implement `artifact_repackage` macro

### DIFF
--- a/distribution/artifact/artifact-repackage.sh
+++ b/distribution/artifact/artifact-repackage.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+in=$1
+out=$(pwd)/$2
+files_to_keep=$3
+
+dest=$(mktemp -d)
+
+if [[ "$(file --dereference --mime-type --brief $in)" == "application/zip" ]]; then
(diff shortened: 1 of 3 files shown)


distribution/artifact/artifact-repackage.sh
```diff
@@ -0,0 +1,23 @@
...
+else
+   tar -xzf "$in" -C "$dest" &>/dev/null
+fi
+
+f=("$dest"/*)
+mv "$dest"/*/* "$dest"
+rmdir "${f[@]}"
+cd $dest || exit 1
+
+rm -f "$files_to_remove"
```
lriuui0x0, line 20: - I think we should be able to remove a folder.
- Isn't `grep -v` solution quite simple? Why do you think it's not reliable?

vmax: @lriuui0x0 addressed in https://github.com/graknlabs/dependencies/pull/260/commits/ef0a58aef5b7d7cc381b3874ab0071cade543e0e

lriuui0x0 (approved):

PR #283, opened 2021-05-13
Introduce a tool to rewrite ANTLR grammars

vmax:

Help advance TypeQL adoption by deploying `typeql-grammar` packages for major languages. For some languages source grammar file needs to be rewritten due to clashes between the names the grammar defines and the names native to the language's runtime. In Python case, such clashing names include `filter` and `type` which would be postfixed with `_` and become `filter_` and `type_`, respectively.

* Bump dependency on `rules_antlr` to a latest release
* Introduce a generic tool to remap ANTLR grammars (`@vaticle_dependencies//builder/antlr:grammar-adapter`)
* Introduce a wrapper rule `python_grammar_adapter` which handles grammar conversion for Python specifically

Intended usage:

```
load("@vaticle_dependencies//builder/antlr:rules.bzl", "python_grammar_adapter")

python_grammar_adapter(
    name = "python-grammar",
    input = "TypeQL.g4",
    output = "TypeQLPython.g4",
)
```

Takes `TypeQL.g4` and produces `TypeQLPython.g4` with the altered grammar name; `type` changed to `type_` and `filter` changed to `filter_`

### DIFF
--- a/builder/antlr/GrammarAdapter.kt
+++ b/builder/antlr/GrammarAdapter.kt
@@ -9,8 +9,8 @@ import java.util.concurrent.Callable
 import kotlin.system.exitProcess
 
 
-@Command(name = "grammar-modifier", mixinStandardHelpOptions = true)
-class GrammarModifier : Callable<Unit> {
+@Command(name = "grammar-adapter", mixinStandardHelpOptions = true)
+class GrammarAdapter : Callable<Unit> {
 
     @Option(names = ["--in"], required = true)
     lateinit var inputFile: File
@@ -50,4 +50,4 @@ class GrammarModifier : Callable<Unit> {
     }
 }
 
-fun main(args: Array<String>): Unit = exitProcess(CommandLine(GrammarModifier()).execute(*args))
+fun main(args: Array<String>): Unit = exitProcess(CommandLine(GrammarAdapter()).execute(*args))
--- a/builder/antlr/GrammarAdapter.kt
+++ b/builder/antlr/GrammarAdapter.kt
@@ -18,7 +18,7 @@ class GrammarAdapter : Callable<Unit> {
     @Option(names = ["--out"], required = true)
     lateinit var outputFile: File
 
-    @Option(names = ["--keyword"], required = true)
+    @Option(names = ["--adapt-keyword"], required = true)
     lateinit var keywords: Array<String>
 
     var patterns = ArrayList<Regex>()
--- a/builder/antlr/rules.bzl
+++ b/builder/antlr/rules.bzl
@@ -6,6 +6,6 @@ def python_grammar_adapter(
         name = name,
         srcs = [input],
         outs = [output],
-        cmd_bash = "$(location @vaticle_dependencies//builder/antlr:grammar-adapter) --in $< --out $@ --keyword type --keyword filter",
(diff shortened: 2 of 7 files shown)


lolski:
```
genrule(
    name = "python-grammar",
    srcs = ["TypeQL.g4"],
    outs = ["TypeQLPython.g4"],
    cmd_bash = "$(location @vaticle_dependencies//builder/antlr:grammar-converter) --in $< --out $@ --keyword filter --keyword type",
    tools = ["@vaticle_dependencies//builder/antlr:grammar-converter"],
)
```

Is this how it's meant to be used @vmax ? I think that's not user friendly enough. Let's make a macro for the language that we want to support:

```
python_grammar_converter(...):
  genrule(
    cmd_bash = @vaticle_dependencies//builder/antlr:grammar-converter ...
  )

nodejs_grammar_converter(...):
  genrule(
    cmd_bash = @vaticle_dependencies//builder/antlr:grammar-converter ...
  )
```

This way user can just use `python_grammar_converter` and `nodejs_grammar_converter` rather than the `genrule` directly.

builder/antlr/rules.bzl
```diff
@@ -0,0 +1,11 @@
+def python_grammar_adapter(
+        name,
+        input,
+        output):
```
lolski, line 4: Can we make `output` optional? It doesn't need to be specified by the user most of the time. We can derive it using the input:

```
native.genrule(
  name = name,
  srcs = [input],
  outs = ['python_' + input]
  ...
)
```

Is my opinion correct @vmax ?

If that's the case, the only time a user needs to set the `output` is if they have multiple targets where each happens to generate a file with the same name. In that rare case, the user can override the default output to avoid the clash.

vmax: @lolski it's actually `TypeQL<Language>.g4`, not `python_TypeQL`.

Thing is, we want users of the rule to be very intentional in how they name the output file - it affects how files, modules and classes are named in the generated code. If we generate the name for the user, we will lose the explicitness part.

lolski: I see, got it @vmax .

builder/antlr/GrammarAdapter.kt
```diff
@@ -0,0 +1,53 @@
...
+@Command(name = "grammar-adapter", mixinStandardHelpOptions = true)
+class GrammarAdapter : Callable<Unit> {
+
+    @Option(names = ["--in"], required = true)
+    lateinit var inputFile: File
+
+    @Option(names = ["--out"], required = true)
+    lateinit var outputFile: File
+
+    @Option(names = ["--keyword"], required = true)
```
lolski, line 21: Can we change it to `--adapt-keyword`?

When put together with the other option, `--keyword` is not clear:

```
--input in.g4 --output  out.g4 --keyword type --keyword filter
```

this is clearer

```
--input in.g4 --output  out.g4 --adapt-keyword type --adapt-keyword filter
```

vmax: @lolski addressed in https://github.com/vaticle/dependencies/pull/283/commits/594fd64558c895bae69d7b57cec6ef0d9d8f15b6

lolski (approved):

PR #299, opened 2021-08-11
Upgrade Kotlin to 1.5 and support Compose

alexjpwalker:

Previously we were using the latest stable release of `rules_kotlin` which used Kotlin 1.3. But in order to use the emerging technology JetBrains Compose, we needed 1.5. So we forked and upgraded `rules_kotlin` and it now uses Kotlin 1.5.

- Upgrade `rules_kotlin` to use Kotlin 1.5
- Add `builder/compose:deps` which downloads the Compose Compiler Kotlin plugin and makes it available as a `filegroup`
- Add ability to skip the SHA check when downloading Maven dependencies (because the Compose Compiler doesn't specify a SHA)
- Add `runtime_deps`, `mac_runtime_deps`, `linux_runtime_deps` and `windows_runtime_deps` to `native_libraries_deps` (because Compose has a platform-native runtime dependency on `org.jetbrains.skiko:skiko-jvm-runtime-xxx`)


library/maven/rules.bzl
```diff
@@ -18,10 +18,13 @@ def maven(artifacts_org, artifacts_repo={}):
...
             "https://repo.vaticle.com/repository/maven",
             "https://repo.vaticle.com/repository/maven-snapshot",
+            "https://dl.google.com/dl/android/maven2",
+            "https://maven.pkg.jetbrains.space/public/p/compose/dev",
         ],
         strict_visibility = True,
         version_conflict_policy = "pinned",
-        fetch_sources = True
+        fetch_sources = True,
+        fail_on_missing_checksum = fail_on_missing_checksum
```
alexjpwalker, line 27: By default, `maven` looks for the presence of a `SHA` or `MD5` hash, but many Compose dependencies don't have one - for example:
```
checksum errors: SHA-1: not found: https://maven.pkg.jetbrains.space/public/p/compose/dev/org/jetbrains/compose/ui/ui-text-desktop/1.0.0-alpha3/ui-text-desktop-1.0.0-alpha3.pom.sha1, MD5: not found: https://maven.pkg.jetbrains.space/public/p/compose/dev/org/jetbrains/compose/ui/ui-text-desktop/1.0.0-alpha3/ui-text-desktop-1.0.0-alpha3.pom.md5
```
In order to fetch these dependencies we have to disable `fail_on_missing_checksum` for the Studio repo.

library/maven/rules.bzl
```diff
@@ -18,10 +18,13 @@ def maven(artifacts_org, artifacts_repo={}):
             "https://repo1.maven.org/maven2",
             "https://repo.vaticle.com/repository/maven",
             "https://repo.vaticle.com/repository/maven-snapshot",
+            "https://dl.google.com/dl/android/maven2",
+            "https://maven.pkg.jetbrains.space/public/p/compose/dev",
```
alexjpwalker, line 22: These Maven repositories contain Compose dependencies.

library/maven/artifacts.bzl
```diff
@@ -227,6 +228,28 @@ artifacts = {
...
+    "org.jetbrains.compose.ui:ui-graphics-desktop": "1.0.0-alpha3",
+    "org.jetbrains.compose.ui:ui-text-desktop": "1.0.0-alpha3",
+    "org.jetbrains.compose.ui:ui-tooling-preview-desktop": "1.0.0-alpha3",
+    "org.jetbrains.compose.ui:ui-unit-desktop": "1.0.0-alpha3",
+    "org.jetbrains.compose.ui:ui-util-desktop": "1.0.0-alpha3",
+    "org.jetbrains.skiko:skiko-jvm": "0.3.9",
+    "org.jetbrains.skiko:skiko-jvm-runtime-linux-x64": "0.3.9",
+    "org.jetbrains.skiko:skiko-jvm-runtime-macos-x64": "0.3.9",
+    "org.jetbrains.skiko:skiko-jvm-runtime-windows-x64": "0.3.9",
+    "org.jetbrains:annotations": "13.0",
```
alexjpwalker, line 252: I am not a fan of this massive dependency list, but removing them does seem to cause issues: see https://github.com/vaticle/typedb-workbase/issues/453

lolski: I think `"org.jetbrains:annotations": "13.0"` should be at the top of the list of the newly added deps, since we want to keep the file in alphabetical / lexicographical order.

lolski (approved):

PR #318, opened 2021-11-05
Support generating ANTLR Rust code

vmax:

Allow generating Rust code in ANTLR

* Upgrade `rules_antlr` to our fork with support for Rust codegen
* Load `antlr-rust` library from Crates.io
* Add `rust_grammar_adapter` to modify grammar to avoid using `type` (which is a keyword in Rust)


builder/antlr/deps.bzl
```diff
@@ -3,8 +3,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def deps():
     git_repository(
         name = "rules_antlr",
-        remote = "https://github.com/marcohu/rules_antlr",
-        tag = "0.5.0"
+        remote = "https://github.com/vaticle/rules_antlr",
```
vmax, line 6: Until https://github.com/marcohu/rules_antlr/pull/18 is merged, we need to use our own fork with Rust support

library/crates/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+# Mandatory (or Cargo tooling is unhappy)
```
lolski, line 1: Is the `toml` file needed just to depend on `antlr-rust`, @vmax ?

library/crates/remote/BUILD.better_any-0.1.1.bazel
```diff
@@ -0,0 +1,62 @@
+"""
```
lolski, line 1: why do we have so many generated Bazel files? what triggers the auto-creation of these auto-generated files @vmax ?

library/crates/update.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+
+# Script for regenerating BUILD files after Cargo.toml update
+
+set -ex
+
+[[ $(readlink $0) ]] && path=$(readlink $0) || path=$0
+CARGO_TOML_HOME=$(cd "$(dirname "${path}")" && pwd -P)
+
+pushd "$CARGO_TOML_HOME" > /dev/null
```
lolski, line 26: I think the variable name should be `HOME`, because we're setting the "home directory" for this script right?

vmax: @lolski in UNIX systems, `HOME` is a reserved name with a well-understood meaning "home directory of the user". 

What we're really setting here is the _current_ directory for `cargo raze` to operate in - which has to be same directory as the one `Cargo.toml` resides in.

For reference, in `bazel-deps`'s `update.sh` it was called `GRAKN_CORE_HOME` because it pointed to the root of the git repo.

vmax: @lolski addressed in https://github.com/vaticle/dependencies/pull/318/commits/94a192c1602db679d8ec1162e4906539f9f2f08b

lolski (approved):

PR #340, opened 2021-12-22
Update create-notes to escape notes and use time-ordering

flyingsilverfin:

We fix release-note creation by escaping the notes inserted into the release template. The create-notes process used a regex to perform string replacement, which was not treating the replacement value as a literal string, but a regex as well, leading to errors. We also reverse the commits before generating the notes, so we get most-recent-first in the generated github release.

* escape the regex replacement string (ie. the notes markdown)
* reverse the order of commits before generating the release notes, so we get most recent commits first rather than last

### DIFF
--- a/tool/release/createnotes/CreateNotes.kt
+++ b/tool/release/createnotes/CreateNotes.kt
@@ -24,6 +24,7 @@ package com.vaticle.dependencies.tool.release.createnotes
 import java.nio.file.Path
 import java.nio.file.Paths
 import kotlin.io.path.notExists
+import kotlin.text.Regex.Companion.escapeReplacement
 
 fun main(args: Array<String>) {
     val bazelWorkspaceDir = Paths.get(getEnv("BUILD_WORKSPACE_DIRECTORY"))
@@ -50,6 +51,6 @@ private fun writeNotesMd(notes: List<Note>, releaseTemplateFile: Path) {
     val template = releaseTemplateFile.toFile().readText()
     if (!template.matches(".*${Constant.releaseTemplateRegex.pattern}.*".toRegex(RegexOption.DOT_MATCHES_ALL)))
         throw RuntimeException("The release-template does not contain the '${Constant.releaseTemplateRegex}' placeholder")
-    val markdown = template.replace(Constant.releaseTemplateRegex, Note.toMarkdown(notes))
+    val markdown = template.replace(Constant.releaseTemplateRegex, escapeReplacement(Note.toMarkdown(notes)))
     releaseTemplateFile.toFile().writeText(markdown)
 }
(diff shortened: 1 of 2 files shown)


flyingsilverfin:
Merging ahead of time to proceed with release, for @lolski  to review when he as time.

tool/release/createnotes/Note.kt
```diff
@@ -75,7 +75,7 @@ class Note {
                 if (line.startsWith("##")) {
                     header += 1
                 } else if (header == 1) {
-                    goal.append(line)
+                    goal.append(line).append("\n")
```
flyingsilverfin, line 78: we were losing newlines, which ruined formatted PR goals

tool/release/createnotes/CreateNotes.kt
```diff
@@ -50,6 +51,6 @@ private fun writeNotesMd(notes: List<Note>, releaseTemplateFile: Path) {
     val template = releaseTemplateFile.toFile().readText()
     if (!template.matches(".*${Constant.releaseTemplateRegex.pattern}.*".toRegex(RegexOption.DOT_MATCHES_ALL)))
         throw RuntimeException("The release-template does not contain the '${Constant.releaseTemplateRegex}' placeholder")
-    val markdown = template.replace(Constant.releaseTemplateRegex, Note.toMarkdown(notes))
+    val markdown = template.replace(Constant.releaseTemplateRegex, escapeReplacement(Note.toMarkdown(notes)))
```
flyingsilverfin, line 54: the `replace(regex, string)` actually treats the replacement string as a sort of regex itself, so we escape it to get the literal string value into the template

tool/release/createnotes/Note.kt
```diff
@@ -120,7 +120,7 @@ ${others.map(Note::toMarkdown).joinToString("\n")}
 }
 
 fun collectNotes(org: String, repo: String, commits: List<String>, githubToken: String): List<Note> {
-    return commits.flatMap { commit ->
+     return commits.reversed().flatMap { commit ->
```
flyingsilverfin, line 123: I did this here, but pushed a commit to master to do this in a better place as a follow up. The commits appear to be in time-order, so reversing them puts most-recent first, which is how we want the release-notes made

lolski: Hmm why would reverse order be better?

flyingsilverfin: because we always list our release notes with most recent first, not most recent last (well, at least this is how we used to do it!) I found it more intuitive for some reason

tool/release/createnotes/Note.kt
lolski: @flyingsilverfin on second thought, although I agree with the importance of adding a newline for formatting purpose, it shouldn't be the responsibility of this method. As the name should imply, `getPRGoal` should be responsible for one thing: extracting the "goal" section ouf of a PR description and without any modification.

Which function should add the newline so that we can keep `getPRGoal` from violating its contract?

PR #341, opened 2022-01-06
Fix incorrect packages and Bazel targets

lolski:

We fixed the `release_validate_deps` rule which was broken due to a recent refactor.

Updated the package and Bazel targets that we forgot to update in the refactor.

### DIFF
--- a/tool/release/deps/rules.bzl
+++ b/tool/release/deps/rules.bzl
@@ -48,7 +48,7 @@ _release_validate_deps_script_test = rule(
         ),
         "_release_validate_deps_template": attr.label(
             allow_single_file=True,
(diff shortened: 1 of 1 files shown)


tool/release/deps/rules.bzl
```diff
@@ -48,7 +48,7 @@ _release_validate_deps_script_test = rule(
         ),
```
lolski, line 48: Update packages and Bazel targets which we forgot to update in https://github.com/vaticle/dependencies/pull/327.

flyingsilverfin (approved):

flyingsilverfin:
presumably you searched for any other occurrences too!

PR #344, opened 2022-01-17
Update bazel-distribution

alexjpwalker:

We updated bazel-distribution to the latest version

Update bazel-distribution


lolski (approved):

lolski:
@alexjpwalker I think the `dependencies` tag was automatically created by Github's Dependabot and we shouldn't use it. In this case, for example, the `type:build` tag is most appropriate.

PR #351, opened 2022-04-26
Make 'release version bump' tool push to the remote Git repo

alexjpwalker:

`//release/version:bump` no longer creates a GitHub milestone, and now pushes to the remote Git repo.

Previously, `//release/version:bump` behaved oddly, modifying the `VERSION` file but not committing the result to any repo. So we rebuilt it. The behaviour of the rule is:

1. Increment the number in the `VERSION` file
2. `git commit`
3. `git push`

### DIFF
--- a/tool/release/version/BUILD
+++ b/tool/release/version/BUILD
@@ -22,7 +22,9 @@ kt_jvm_binary(
     srcs = ["Bump.kt"],
     main_class = "com.vaticle.dependencies.tool.release.version.BumpKt",
     deps = [
-        "@maven//:com_eclipsesource_minimal_json_minimal_json",
-        "@maven//:com_google_http_client_google_http_client",
+        "@vaticle_bazel_distribution//common",
+        "@vaticle_bazel_distribution//common/shell",
+
+        "@maven//:org_zeroturnaround_zt_exec",
     ],
 )
--- a/tool/release/version/BUILD
+++ b/tool/release/version/BUILD
@@ -17,6 +17,8 @@
 
 load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_binary")
 
+package(default_visibility = ["//visibility:public"])
+
 kt_jvm_binary(
     name = "bump",
     srcs = ["Bump.kt"],
--- a/tool/release/version/Bump.kt
+++ b/tool/release/version/Bump.kt
@@ -18,6 +18,7 @@ fun main() {
         ?: throw RuntimeException("GRABL_BRANCH environment variable is not set")
     val workspacePath = Paths.get(workspaceDirectory)
 
+    configureGitCredentials()
     shell.execute(listOf("git", "checkout", branchName), baseDir = workspacePath)
 
     val versionFile = workspacePath.resolve("VERSION")
@@ -29,6 +30,16 @@ fun main() {
     Files.write(versionFile, newVersion.toByteArray())
 
     gitCommitAndPush(workspacePath, newVersion)
+}
+
+fun configureGitCredentials() {
+    val gitUsername = System.getenv("GIT_USERNAME")
+        ?: throw RuntimeException("GIT_USERNAME environment variable is not set")
+    val gitEmail = System.getenv("GIT_EMAIL")
+        ?: throw RuntimeException("GIT_EMAIL environment variable is not set")
+
+    shell.execute(listOf("git", "config", "--global", "user.name", gitUsername))
+    shell.execute(listOf("git", "config", "--global", "user.email", gitEmail))
 }
 
 fun bumpVersion(version: String): String {
@@ -57,7 +68,7 @@ fun bumpVersion(version: String): String {
 
 fun gitCommitAndPush(workspacePath: Path, newVersion: String) {
     shell.execute(listOf("git", "add", "VERSION"), baseDir = workspacePath)
-    shell.execute(listOf("git", "commit", "-m", "\"Bump version number to $newVersion\""), baseDir = workspacePath)
+    shell.execute(listOf("git", "commit", "-m", "Bump version number to $newVersion"), baseDir = workspacePath)
     val maxRetries = 3
     var retryCount = 0
     while (retryCount < maxRetries) {
--- a/tool/release/version/Bump.kt
+++ b/tool/release/version/Bump.kt
@@ -76,7 +76,7 @@ fun bumpVersion(version: String): String {
 
 fun getRemoteURL(workspacePath: Path, gitToken: String): String {
     val gitRemoteOutput = shell.execute(listOf("git", "remote", "-v"), baseDir = workspacePath).outputString()
-    val regex = Regex("git@github\\.com:(\\w+)/(\\w+)\\.git")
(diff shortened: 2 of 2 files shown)


tool/release/version/BUILD
```diff
@@ -22,7 +22,9 @@ kt_jvm_binary(
     srcs = ["Bump.kt"],
```
alexjpwalker, line 22: We no longer need to call the GitHub API, but we do need to run shell commands.

tool/release/version/Bump.kt
```diff
@@ -1,26 +1,17 @@
 package com.vaticle.dependencies.tool.release.version
```
alexjpwalker, line 1: We've kept the logic for incrementing the version number, but taken out everything else. The steps are now:
1. increment number in `VERSION` file
2. commit `VERSION` file and push

tool/release/version/Bump.kt
```diff
@@ -35,34 +26,41 @@ fun bumpVersion(version: String): String {
...
     return versionComponents.joinToString(".")
 }
 
+fun gitCommitAndPush(workspacePath: Path, newVersion: String) {
+    val shell = Shell(logger)
+    shell.execute(listOf("git", "add", "VERSION"), baseDir = workspacePath)
+    shell.execute(listOf("git", "commit", "-m", "\"Bump version number to $newVersion\""), baseDir = workspacePath)
+    val maxRetries = 3
+    var retryCount = 0
+    while (retryCount < maxRetries) {
```
alexjpwalker, line 42: We pull first to ensure that the `push` succeeds. If there are merge conflicts, my understanding is that `git` will return a non-zero exit code.

tool/release/version/Bump.kt
```diff
@@ -35,34 +27,44 @@ fun bumpVersion(version: String): String {
...
-    val repoOwner = System.getenv("GRABL_OWNER")
-            ?: throw RuntimeException("GRABL_OWNER environment variable is not set")
-    val repoName = System.getenv("GRABL_REPO")
-            ?: throw RuntimeException("GRABL_REPO environment variable is not set")
+    val workspacePath = Paths.get(workspaceDirectory)
+
+    shell.execute(listOf("git", "checkout", "master"), baseDir = workspacePath)
 
-    val versionFile = Paths.get(workspaceDirectory, "VERSION")
+    val versionFile = workspacePath.resolve("VERSION")
```
lolski, line 62: While we can assume most repos will have a `VERSION` file, it's still nice to do an assertion and throw a meaningful error message when it doesn't:

```
val versionFile = workspacePath.resolve("VERSION")
if (!versionFile.exists()) throw RuntimeException("...")
```

tool/release/version/Bump.kt
lolski: Interesting, I suppose it is possible.

What if we simplify the approach to simply always pulling before making the commit:

```
fun gitCommitAndPush(workspacePath: Path, newVersion: String) {
    shell.execute(listOf("git", "pull"), baseDir = workspacePath)
    shell.execute(listOf("git", "add", "VERSION"), baseDir = workspacePath)
    shell.execute(listOf("git", "commit", "-m", "\"Bump version number to $newVersion\""), baseDir = workspacePath)
    shell.execute(listOf("git", "push"), baseDir = workspacePath)
}
```

alexjpwalker:
> Should we reorder the methods such that `main` is at the top of the file?

We generally put `main` at the bottom of the file in build scripts, eg `JarAssembler` and `PomGenerator` from `bazel-distribution`, such that the things `main` depends on are defined first, and then `main` is defined afterwards.

tool/release/version/Bump.kt
alexjpwalker: One of the possible failure scenarios is that someone else pushes between the `git pull` and `git push`, causing the push to fail. So I think a retry loop is the most reliable way to handle this admittedly rare case.

tool/release/version/Bump.kt
```diff
@@ -35,34 +27,44 @@ fun bumpVersion(version: String): String {
...
             ?: throw RuntimeException("Not running from within Bazel workspace")
-    val githubToken = System.getenv("GITHUB_TOKEN")
-            ?: throw RuntimeException("GITHUB_TOKEN environment variable is not set")
-    val repoOwner = System.getenv("GRABL_OWNER")
-            ?: throw RuntimeException("GRABL_OWNER environment variable is not set")
-    val repoName = System.getenv("GRABL_REPO")
-            ?: throw RuntimeException("GRABL_REPO environment variable is not set")
+    val workspacePath = Paths.get(workspaceDirectory)
+
+    shell.execute(listOf("git", "checkout", "master"), baseDir = workspacePath)
```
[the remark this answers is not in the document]
alexjpwalker: Grabl currently checks out `$GRABL_COMMIT`:
```
Agent pid ****
Added user key of ****
bazel-setup-rbe 
No RBE credential found. Bazel will be executed locally without RBE support.
Configuring Python...
git clone git@github.com:$GRABL_OWNER/$GRABL_REPO.git 
Cloning into 'web-main'...
Warning: Permanently added the RSA host key for IP address '****' to the list of known hosts.
cd $GRABL_REPO 
git checkout $GRABL_COMMIT 

Note: switching to 'a9ef54f232d5eb85b8434fda55588411e436a009'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c 

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at a9ef54f2 "Bump version number to 2.0.1"
bazel build //... 
```

tool/release/version/Bump.kt
lolski: Hmm I think that's rare enough to justify the need of handling that. Because if we want to be that robust, then for the sake of consistency we will have to be just as robust in every other Bazel target in the repo, mainly for handling network and disk issues, ending up with lots of retry loops.

Therefore for the benefit of code simplicity I'd say we go with the simplified approach. What do you think?

tool/release/version/Bump.kt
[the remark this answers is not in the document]
lolski: Can you check if we have the variable `GRABL_BRANCH` defined in jobs?

tool/release/version/Bump.kt
alexjpwalker: We decided to keep the retry logic because someone else pushing is an "expected" error (despite being very rare in this case). We also improved the logic, now checking the exit value of `git push` instead of just allowing it to throw an exception and catching it.

alexjpwalker:
> Should we reorder the methods such that `main` is at the top of the file?

We decided to move `main` to the top of the file on the grounds that it is the most natural layout. For standalone scripts like this one, and `JarAssembler` and more, it makes sense for the entry point to be the first thing the user reads. We will move `main` methods to the top of those files as well as we come across them and need to edit them for whatever purpose.

lolski (approved):

PR #375, opened 2022-10-14
Rust IDE sync tool

alexjpwalker:

We've added tooling to "sync" a Bazel project that uses Rust via `@rules_rust` such that IDEs, like IntelliJ and CLion, are able to support its development - providing IntelliSense and so on.

When we began work on the IntelliJ Bazel + Rust plugins 9 months ago, we aimed to create fully-functional IDE tooling that was fully faithful to Bazel's methodology. Due to the substantial difficulty of getting things to work (e.g: unable to debug a Bazel **rust_test**), we changed strategy. Taking bits and pieces of the work we'd done on those plugins (most notably the Bazel Plugin's `Cargo.toml` generation), we no longer centre our development workflow around Bazel targets, but rather add tooling to generate Cargo projects from our Bazel projects automatically. This allows us to leverage the full power of the IntelliJ Rust plugin to develop seamlessly in Rust.

### How it works

We added a Kotlin binary (`@vaticle_dependencies//ide/rust:sync`) that, when run in a Bazel project that uses `rules_rust`, generates the `Cargo.toml` files that define a Cargo project. It runs the following steps:

1. Validate the project Rust targets by doing `bazel build --build=false` on the targets whose names start with `rust_`. This verifies correctness of the Rust `BUILD` targets, as things like incorrect dependency declarations can derail the sync tool.
2. In the Bazel workspace directory, `bazel build` the Rust targets. This achieves two purposes: firstly, it loads the Rust toolchain and stdlib that IntelliJ and CLion use; secondly, it loads the external `@vaticle` Rust repos so we can sync them too. Errors at this stage are ignored, as they typically represent errors in our Rust code, and we want the user to be able to sync the project in that scenario.
3. In the current workspace, and in all external `@vaticle` repositories* that Rust targets depend on:
    1. Run the **Rust IDE Sync Info Aspect**, outputting metadata in the form of `.ide-sync.properties` files in `bazel-bin`. Using an Aspect allows us to easily retrieve crate names and versions of the entire dependency tree.
    2. Run the **Cargo Project Generator**, a Kotlin/JVM binary that recursively searches `bazel-bin` for IDE metadata, and transforms them to `Cargo.toml` files in the project directory. The Aspect alone can't do this, because a) it can only write to `bazel-bin` and b) Cargo and Bazel differ in the way test dependencies are declared; a Cargo library needs to know about its tests, but a Bazel Rust test target needs to know about its library.

\* _"Repository" here refers to a [Bazel repository](https://bazel.build/extending/repo), such as `@vaticle_typeql`._

#### Why sync external `@vaticle` repositories?

Whenever a Rust target in our current project depends on an external `@vaticle` Rust target, its generated `Cargo.toml` contains a reference to that external target. Any dependency in a `Cargo.toml` must itself have a `Cargo.toml`; hence, we must sync all external `@vaticle` dependencies as well as the current project.

Why use "starts with `@vaticle`" as the filter, rather than something more generic? Because it keeps the logic very simple. If, in the future, we depend on non-`vaticle` Bazel Rust projects, we'll need to update the filter.

### Usage

In each Bazel project that uses Rust, add the following executable script, which we name `dependencies/ide/sync.sh`:
```sh
bazel run @vaticle_dependencies//ide/rust:sync
```
After "importing the Bazel project", immediately run `dependencies/ide/sync.sh`. Once `sync.sh` completes, the Cargo project will be generated.

Add `target`, `Cargo.lock` and `Cargo.toml` to `.gitignore`. (`target` is generated by a Cargo build.)

The Sync tool supports the `--verbose` (`-v`) command-line flag to enable verbose output (`bazel run @vaticle_dependencies//ide/rust:sync -- --verbose`)

### Compatibility requirements

Currently, the sync tool only works with `@vaticle` Bazel dependencies (see "Why sync external `@vaticle` repositories" above). If we want to depend on Bazel projects from a different org, we'll need to update the relevant code.

We also assume that the Rust support in Bazel is delivered via `@rules_rust`. This is a fair assumption; it is the standard library for Rust rules.

We assume that external crates are loaded via `cargo raze`. This is currently true for all crates declared in `@vaticle_dependencies`. If we were to refactor in the future to use, say, `crate_universe`, we would inevitably have to update the sync tool as well.

It's also possible that API changes and refactors in `rules_rust` will result in further changes needing to be made to our sync tool. For example, a build script is currently detected on the basis of its label being `_build_script_run`. As this is an internal/private symbol in `rules_rust`, it's quite likely to change in the future.


grabl:
### PR Review Checklist

Do not edit the content of this comment. The PR reviewer should simply update this comment by ticking each review item below, as they get completed.

----
**Code**

- [ ] Packages, classes, and methods have a single domain of responsibility.
- [ ] Packages, classes, and methods are grouped into cohesive and consistent domain model.
- [ ] The code is canonical and the minimum required to achieve the goal.
- [ ] Modules, libraries, and APIs are easy to use, robust (foolproof and not errorprone), and tested.
- [ ] Logic and naming has clear narrative that communicates the accurate intent and responsibility of each module (e.g. method, class, etc.).
- [ ] The code is algorithmically efficient and scalable for the whole application.

**Architecture**

- [ ] Any required refactoring is completed, and the architecture does not introduce technical debt incidentally.
- [ ] Any required build and release automations are updated and/or implemented.
- [ ] Any new components follows a consistent style with respect to the pre-existing codebase.
- [ ] The architecture intuitively reflects the application domain, and is easy to understand.
- [ ] The architecture has a well-defined hierarchy of encapsulated components.
- [ ] The architecture is extensible and scalable.

builder/rust/cargo/sync_aspect.bzl
```diff
@@ -0,0 +1,162 @@
+#
```
alexjpwalker, line 1: This is the Cargo Sync Aspect, which propagates through the Rust dependency tree via the `deps`, `proc_macro_deps` and `crate` attributes (which exist on some or all of `[rust_binary, rust_library, rust_test]`). It aggregates the crate names and versions (and Cargo `features`) of the Rust targets that it traverses, providing that information to the targets higher up in the dependency tree that depend on those targets.

In the end, we use that information to produce Cargo Sync metadata in the form of `.cargo-sync.properties` files in `bazel-bin` (e.g: `typeql_lang.cargo-sync.properties`), which does nothing by itself, but the Project Structure Builder analyses them and uses them to generate Cargo project structure in the form of `Cargo.toml` files.

tool/common/deps.bzl
```diff
@@ -8,7 +8,9 @@ def vaticle_dependencies_ci_pip():
```
alexjpwalker, line 8: Added TOML library to handle `Cargo.toml`, same one we use in `bazel-distribution` and elsewhere.

lolski:
Is it possible for us to format the generated `Cargo.toml` files? They tend to contain superfluous whitespaces and the formatting isn't so nice.

builder/rust/cargo/generate-project.sh
```diff
@@ -0,0 +1,24 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+set -x
```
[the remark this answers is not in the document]
alexjpwalker: To be honest, at that point we're into the realm of parsing command-line parameters in Bash, which is not fun and plagued with edge cases.

I'd be inclined to just remove `set -x` (replacing it with `set -e`).

alexjpwalker:
We had previously discussed renaming `sync_aspect.bzl` to `metadata_aspect.bzl`. However:

- `metadata` doesn't carry very much meaning.
- The aspect is used specifically when we run `sync`; and the name `sync` is not anywhere in the package structure. Therefore, it actually makes sense to continue calling the file `sync_aspect`, and rename the aspect itself also to `sync_aspect` (taking `cargo_` out of the aspect name)

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,115 @@
...
+        logger.debug { "Cleaning up old sync info under $repoPath" }
+        val bazelBin = File(shell.execute(listOf(BAZEL, INFO, "bazel-bin"), repoPath).outputString().trim())
+        bazelBin.listFilesRecursively().filter { it.name.endsWith(".ide-sync.properties") }.forEach { it.delete() }
+    }
+
+    private fun runSyncInfoAspect(repoPath: Path) {
+        val rustTargets = rustTargets(repoPath)
+        shell.execute(listOf(BAZEL, BUILD) + rustTargets + listOf(
+            "--aspects",
+            "@vaticle_dependencies//ide/rust:sync_aspect.bzl%rust_ide_sync_aspect",
```
lolski, line 104: Again, these may break when we refactor the repo. Let's see if we can test them or gather them into a single "constant" or "config" object at least.

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,115 @@
...
+    private fun rustTargets(repoPath: Path): List<String> {
+        return shell.execute(listOf(BAZEL, QUERY, "kind(rust_*, //...)"), repoPath)
+            .outputString().split("\n").filter { it.isNotBlank() }
+    }
+
+    private fun rustRepoPaths(): List<Path> {
+        // e.g: [/Users/root/workspace/typedb-client-rust, /private/var/_bazel_root_/123abc/external/vaticle_typedb_protocol]
+        return listOf(buildWorkspaceDir) + shell.execute(listOf(BAZEL, QUERY, "kind(rust_*, deps(//...))", "--output=package"), buildWorkspaceDir)
+            .outputString().split("\n")
+            .filter { it.startsWith("@vaticle_") }
```
lolski, line 81: Paths like these will break when we refactor the repo. Is there any way that we can test them? At the very least, we gather them into a single "constant" or "config" object.

lolski (changes requested):

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,128 @@
...
+    private fun loadRustToolchainAndExternalDeps(rustTargets: List<String>) {
+        shell.execute(command = listOf(BAZEL, BUILD) + rustTargets, baseDir = buildWorkspaceDir, throwOnError = false)
+    }
+
+    private fun rustTargets(repoPath: Path): List<String> {
+        return shell.execute(listOf(BAZEL, QUERY, RUST_TARGETS_QUERY), repoPath)
+            .outputString().split("\n").filter { it.isNotBlank() }
+    }
+
+    private fun rustRepoPaths(): List<Path> {
```
lolski, line 84: This method does feel quite terse, therefore let's make it less so. Also the name can be more descriptive. Something like `vaticleRustRepo()` perhaps?

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private fun validateBuildWorkspace(rustTargets: List<String>) {
+        shell.execute(command = listOf(BAZEL, BUILD) + rustTargets + "--build=false", baseDir = buildWorkspaceDir)
+    }
+
+    private fun loadRustToolchainAndExternalDeps(rustTargets: List<String>) {
+        shell.execute(command = listOf(BAZEL, BUILD) + rustTargets, baseDir = buildWorkspaceDir, throwOnError = false)
+    }
+
+    private fun rustTargets(repoPath: Path): List<String> {
```
lolski, line 79: this method is used by `call()` method and `runSyncInfoAspect()`. Therefore according to our convention it should be placed somewhere in the bottom.

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,128 @@
...
+import java.util.concurrent.Callable
+import kotlin.io.path.Path
+import kotlin.system.exitProcess
+
+fun main(args: Array<String>): Unit = exitProcess(CommandLine(SyncRunner()).execute(*args))
+
+@CommandLine.Command(name = "sync", mixinStandardHelpOptions = true)
+class SyncRunner : Callable<Unit> {
+
+    @CommandLine.Option(names = ["--verbose", "-v"], required = false)
```
lolski, line 53: can some of these variables be private?

lolski (changes requested):

ide/rust/SyncRunner.kt
alexjpwalker: Renamed to `vaticleRustRepoPaths`. What do you mean by "let's make it less so"?

ide/rust/SyncRunner.kt
alexjpwalker: Moved

ide/rust/SyncRunner.kt
alexjpwalker: Yes, they should all be private.

Getting too comfortable in Rust where things are private by default :(

ide/rust/CargoManifestGenerator.kt
```diff
@@ -0,0 +1,322 @@
+/*
```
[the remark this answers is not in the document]
alexjpwalker: Most obvious split would be to put `IDESyncInfo` (currently an inner class of `CargoManifestGenerator`) into its own file, so let's do that.

lolski:
> > Is it possible for us to format the generated `Cargo.toml` files? They tend to contain superfluous whitespaces and the formatting isn't so nice.
> 
> We could do some simple but inelegant replacements, e.g. collapsing multiple whitespaces into single whitespaces (to the string itself, using our code). The formatting options provided by the config library itself are limited because it is a highly generic library which supports other formats such as JSON; it is not that rich in features.

I see. Let's keep it as is if we have no control over it.

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private fun validateBuildWorkspace(rustTargets: List<String>) {
+        shell.execute(command = listOf(BAZEL, BUILD) + rustTargets + "--build=false", baseDir = buildWorkspaceDir)
+    }
+
+    private fun loadRustToolchainAndExternalDeps(rustTargets: List<String>) {
+        shell.execute(command = listOf(BAZEL, BUILD) + rustTargets, baseDir = buildWorkspaceDir, throwOnError = false)
+    }
+
+    private fun vaticleRustRepoPaths(): List<Path> {
```
lolski, line 79: I think we can just call it `vaticleRustRepos` or `vaticleRustRepositories`. It is clear enough that the repositories are represented by the path at which it is located.

ide/rust/SyncRunner.kt
```diff
@@ -0,0 +1,128 @@
...
+        // e.g: [/Users/root/workspace/typedb-client-rust, /private/var/_bazel_root_/123abc/external/vaticle_typedb_protocol]
+        return listOf(buildWorkspaceDir) + shell.execute(listOf(BAZEL, QUERY, RUST_TARGETS_DEPS_QUERY, "--output=package"), buildWorkspaceDir)
+            .outputString().split("\n")
+            .filter { it.startsWith(VATICLE_REPO_PREFIX) }
+            .map { it.split("@")[1].split("//")[0] }
+            .map { bazelOutputBase.resolve("external").resolve(it) }
+    }
+
+    private fun sync(repoPath: Path) {
```
lolski, line 88: Similar to how we should rename `vaticleRustRepoPaths`, I think we can rename `repoPath` to just `repo` or `repository.`

Reason: It seems that repositories are always represented as path objects anyway, so referring to them as "repo path" feels redundant.

ide/rust/RepoCargoManifestGenerator.kt
```diff
@@ -0,0 +1,325 @@
...
+import com.vaticle.dependencies.ide.rust.RepoCargoManifestGenerator.TargetSyncInfo.Keys.VERSION
+import java.io.File
+import java.io.FileInputStream
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.Properties
+import kotlin.io.path.Path
+
+class RepoCargoManifestGenerator(private val workspaceRoot: File, shell: Shell) {
```
lolski, line 58: For consistency, the field `workspaceRoot` should be named `repository`, don't you think?

lolski (changes requested):

lolski (approved):

ide/rust/BUILD
```diff
@@ -0,0 +1,44 @@
+#
```
alexjpwalker, line 1: `:sync` is the entry point for the tool.

ide/rust/RepoCargoManifestGenerator.kt
```diff
@@ -0,0 +1,325 @@
+/*
```
alexjpwalker, line 1: Given a Bazel repository (either the current project or an external dependency such as `@vaticle_typeql`, `RepoCargoManifestGenerator.generateManifests()` will run the **Rust IDE Sync Info Aspect**, generating IDE metadata files in `bazel-bin`, and then scan the generated files and use them to generate `Cargo.toml` files that constitute a Cargo project.

ide/rust/Syncer.kt
```diff
@@ -0,0 +1,131 @@
+/*
```
alexjpwalker, line 1: `Syncer` is run from a Bazel project that uses Rust. It identifies all the Bazel repositories that have Rust targets that are depended on by the current project, then runs the `RepoCargoManifestGenerator` in each one (plus the current project).

The end result is that the current project and all of its Bazel Rust dependencies will contain `Cargo.toml` files, allowing Cargo to build the project.

ide/rust/sync_info_aspect.bzl
```diff
@@ -0,0 +1,179 @@
+#
```
alexjpwalker, line 1: This is the **Cargo Sync Info Aspect,** which propagates through the Rust dependency tree via the `deps`, `proc_macro_deps` and `crate` attributes (which exist on some or all of `[rust_binary, rust_library, rust_test]`). It aggregates the crate names and versions (and Cargo `features`) of the Rust targets that it traverses, providing that information to the targets higher up in the dependency tree that depend on those targets.

In the end, we use that information to produce Sync Info in the form of `.ide-sync.properties` files in `bazel-bin` (e.g: `typeql_lang.ide-sync.properties`), which does nothing by itself, but the `RepoCargoManifestGenerator` analyses them and uses them to generate Cargo project structure in the form of `Cargo.toml` files.

PR #386, opened 2022-11-16
Add transitive deps of `ortools` that aren't included by default on M1 Macs

alexjpwalker:

We added `core-foundation-sys` and `libc` explicitly as crate dependencies. They are conditional dependencies of `ortools`. Due to a bug, most likely in `cargo raze`, they are not included by default.

`core-foundation-sys` and `libc` are conditional dependencies of the `ortools` crate. The condition is the Cargo condition `cfg('unix')`, which is supposed to represent any Unix system. Due to a bug, most likely in `cargo raze`, they are not correctly pulled in when the machine has `aarch64` architecture.

We'll raise an issue/PR in the relevant upstream repo (most likely `cargo-raze`), but for now, we need a workaround to get our projects building on an M1 Mac. This PR is that workaround. We'll declare the dependencies explicitly, and depend on them explicitly in the targets that require them - most notably `ortools` will explicitly depend on `libc`, and `typeql` will explicitly depend on `libc` and `core-foundation-sys`.

### DIFF
--- a/library/ortools/rust/BUILD
+++ b/library/ortools/rust/BUILD
@@ -26,6 +26,7 @@ rust_library(
     srcs = ["ortools.rs"],
     deps = [
         "@vaticle_dependencies//library/crates:cxx",
+        "@vaticle_dependencies//library/crates:libc",
         ":ortools_bridge",
     ]
 )
--- a/library/crates/BUILD.bazel
+++ b/library/crates/BUILD.bazel
@@ -40,6 +40,15 @@ alias(
 )
 
 alias(
+    name = "core_foundation_sys",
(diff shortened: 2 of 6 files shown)


library/ortools/rust/BUILD
```diff
@@ -26,6 +26,7 @@ rust_library(
     srcs = ["ortools.rs"],
     deps = [
         "@vaticle_dependencies//library/crates:cxx",
+        "@vaticle_dependencies//library/crates:libc",
```
alexjpwalker, line 29: `ortools` depends transitively on `libc`, but this dependency isn't being picked up correctly in an `aarch64` architecture environment, so we now depend on it explicitly.

haikalpribadi (approved):

PR #387, opened 2022-11-16
Support Apple Silicon in `cargo raze` crates

alexjpwalker:

We configured `cargo raze` to generate targets that support Apple Silicon systems.

`cargo raze`'s config file, `Cargo.toml`, defines the list of crates to pull in, as well as the list of platforms to target:
```toml
# The set of targets to generate BUILD rules for.
targets = [
    "x86_64-apple-darwin",
    "x86_64-pc-windows-msvc",
    "x86_64-unknown-linux-gnu",
]
```
We need to add `aarch64-apple-darwin` to support Apple Silicon.

### DIFF
--- a/library/crates/Cargo.toml
+++ b/library/crates/Cargo.toml
@@ -10,12 +10,10 @@ antlr-rust = "=0.3.0-beta"
 axum = "=0.5.15"
 chrono = "=0.4.22"
 crossbeam = "=0.8.2"
-core-foundation-sys = "=0.8.3" # transitive dependency of 'iana_time_zone', included to patch cfg('unix') issue (#385)
 cxx = "=1.0.59"
 derivative = "=2.2.0"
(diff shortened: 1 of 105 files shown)


library/crates/Cargo.toml
```diff
@@ -10,12 +10,10 @@ antlr-rust = "=0.3.0-beta"
 axum = "=0.5.15"
 chrono = "=0.4.22"
 crossbeam = "=0.8.2"
-core-foundation-sys = "=0.8.3" # transitive dependency of 'iana_time_zone', included to patch cfg('unix') issue (#385)
```
alexjpwalker, line 13: We included these dependencies explicitly to patch the issue of them not being installed on an M1 Mac, but now we've actually fixed the root cause of the issue.

haikalpribadi (approved):

PR #391, opened 2022-11-21
Add `rust_tonic_compile` rule

alexjpwalker:

We added the `rust_tonic_compile` rule which compiles `.proto` files into Rust sources that depend on the `tonic` and `prost` crates.

We had an existing Rust binary for compiling proto files into a Tonic library - a Cargo build script in `typedb-protocol`. But Cargo users shouldn't have to run `protoc` on their machine to compile `typedb-client` (which is what happens with a Cargo build script). Rather, the `typedb-protocol` crate should contain the generated Rust gRPC + Protobuf sources.

So we've created a rule, `rust_tonic_compile`, that internally runs a `rust_binary` (similarly to how many of our rules run Kotlin binaries), which uses `tonic_build` to compile `.proto` files into Rust sources, and exposes the outputs. Any `rust_library` can then depend on these generated sources. The API is similar to Java and Python (`java_grpc_compile` and `python_grpc_compile` respectively), and the `antlr` rule from `rules_antlr` behaves similarly, too.

Ideally, we wouldn't need our own rule to do this, however, the Bazel rules repos that contain Rust gRPC + Protobuf rules only support the `grpc` crate, and not the more popular `tonic` crate. Moreover, all of them already have open PRs for adding such functionality:

- https://github.com/rules-proto-grpc/rules_proto_grpc/issues/143
- https://github.com/stackb/rules_proto/pull/201
- https://github.com/bazelbuild/rules_rust/issues/915

Given that our implementation is **very minimal** (~30 lines of code) it should incur a very low maintenance effort.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -21,11 +21,13 @@ workspace(name = "vaticle_dependencies")
 # Load @vaticle_dependencies #
 ################################
 
-# Load //build/rust
+# Load //builder/rust
 load("//builder/rust:deps.bzl", rust_deps = "deps")
 rust_deps()
-load("@rules_rust//rust:repositories.bzl", "rust_repositories")
-rust_repositories(version = "nightly", iso_date = "2021-07-01", edition="2018")
+
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")
+rules_rust_dependencies()
+rust_register_toolchains(edition = "2021", include_rustc_srcs = True)
 
 # Load //builder/python
 load("//builder/python:deps.bzl", python_deps = "deps")
--- a/builder/grpc/rust/BUILD
+++ b/builder/grpc/rust/BUILD
@@ -1,0 +1,33 @@
+#
+# Copyright (C) 2022 Vaticle
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+load("@rules_rust//rust:defs.bzl", "rust_binary")
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+
+rust_binary(
+    name = "compile",
+    srcs = ["compile.rs"],
+    deps = ["@vaticle_dependencies//library/crates:tonic_build"],
+    visibility = ["//visibility:public"]
+)
+
+checkstyle_test(
+    name = "checkstyle",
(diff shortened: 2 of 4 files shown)


builder/grpc/rust/BUILD
```diff
@@ -0,0 +1,33 @@
+#
```
alexjpwalker, line 1: This package contains the `rust_tonic_compile` rule and the `rust_binary` that powers it.

Ideally, we'd get this rule from an external repo such as `rules_rust` or `rules_proto_grpc`, but they don't have `tonic` implementations yet.

We may want to relocate this rule to another repo so it can be used by other Rust proto repos. We may also want to ask the `rules_proto_grpc` maintainers if they'd accept it as a PR. They do have an existing open PR that adds `rust_tonic_compile` (and more), so we'd start by commenting there.

builder/grpc/rust/compile.rs
```diff
@@ -0,0 +1,28 @@
+//
```
alexjpwalker, line 1: We pass in all the `tonic_build` configuration via environment variables, which are set by the `rust_tonic_compile` rule.

builder/grpc/rust/compile.bzl
```diff
@@ -0,0 +1,62 @@
+#
```
alexjpwalker, line 1: The rule implementation is quite straightforward:

1. Define the inputs for the compile script, namely all `.proto` sources and the `protoc` executable.
2. Define the outputs of the compile script. One output is generated for each unique `package` defined in our `.proto` files. Bazel can't (easily!) get these package names, so the user is required to supply them as a parameter.
3. Run the compile script and expose the outputs as source files that any `rust_library` can depend on.

(The `rust_library` is then expected to depend on the `tonic` and `prost` crates, otherwise it will fail to compile. This is in line with Java, where we expect our `java_library` to depend on the correct gRPC + Protobuf JARs.)

WORKSPACE
```diff
@@ -21,11 +21,13 @@ workspace(name = "vaticle_dependencies")
...
-# Load //build/rust
+# Load //builder/rust
 load("//builder/rust:deps.bzl", rust_deps = "deps")
 rust_deps()
-load("@rules_rust//rust:repositories.bzl", "rust_repositories")
-rust_repositories(version = "nightly", iso_date = "2021-07-01", edition="2018")
+
+load("@rules_rust//rust:repositories.bzl", "rules_rust_dependencies", "rust_register_toolchains")
+rules_rust_dependencies()
+rust_register_toolchains(edition = "2021", include_rustc_srcs = True)
```
alexjpwalker, line 30: Update to the latest Rust (fixes build issue in `ortools`, and also is just generally good practice)

builder/grpc/rust/compile.bzl
```diff
@@ -0,0 +1,62 @@
...
+
+    ctx.actions.run(
+        inputs = inputs,
+        outputs = outputs,
+        executable = ctx.executable._compile_script,
+        env = {
+            "OUT_DIR": outputs[0].dirname,
+            "PROTOC": ctx.attr.protoc.files.to_list()[0].path,
+            "PROTOS": ";".join([src.path for src in protos]),
+            "PROTOS_ROOT": ctx.attr.srcs[0][ProtoInfo].proto_source_root,
```
alexjpwalker, line 32: `OUT_DIR` and `PROTOC` are both used internally by the `tonic_build` crate. So, for convenience, we use environment variables to represent the other necessary input arguments as well.

lolski (approved):

PR #392, opened 2022-11-22
Update bazel-distribution

alexjpwalker:


lolski (approved):

.factory/automation.yml
```diff
@@ -22,4 +22,4 @@ build:
     build:
       image: vaticle-ubuntu-22.04
       command: |
-        bazel build //... --deleted_packages=library/rocksdbjni
+        bazel build //...
```
alexjpwalker, line 25: I think we originally used this flag to ignore this package when doing `bazel build //...`, but we now have a better mechanism for that: `tags = ["ignore"]` on all the targets we want to ignore.

And yes, we do want to ignore `library/rocksdbjni`. It takes a **very** long time to build, adding significant CI time.

PR #395, opened 2022-11-23
Add regex crate

dmitrii-ubskii:

We pin the version of the regex crate and make it available through bazel.

### DIFF
--- a/library/crates/BUILD.bazel
+++ b/library/crates/BUILD.bazel
@@ -148,8 +148,26 @@ alias(
 )
 
 alias(
+    name = "regex",
+    actual = "@raze__regex__1_6_0//:regex",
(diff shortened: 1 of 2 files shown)


library/crates/BUILD.bazel
```diff
@@ -156,6 +165,15 @@ alias(
...
 
+alias(
+    name = "rustix",
+    actual = "@raze__rustix__0_35_9//:rustix",
+    tags = [
+        "cargo-raze",
+        "manual",
+    ],
+)
+
```
dmitrii-ubskii, line 176: `update.sh` wasn't apparently run after the last change in the repo, so this is added here.

alexjpwalker (approved):

PR #400, opened 2022-11-28
Rust IDE sync: compile generated sources

alexjpwalker:

The Rust IDE sync tool now compiles generated sources in external repositories.

Previously, `typedb-client-rust` would fail to build using `cargo`, because the sources of `typedb-protocol` - imported as an external Bazel repository - were not automatically compiled by the sync tool. So the generated `Cargo.toml` for that target pointed to a valid `lib.rs` file, but that `lib.rs` file depended on `typedb.protocol.rs`, which did not exist, because it had not been created.

Now, whenever the sync tool finds a Bazel Rust target that contains generated sources, it runs `bazel build` on that target to generate those sources.

### DIFF
--- a/ide/rust/RepoCargoManifestGenerator.kt
+++ b/ide/rust/RepoCargoManifestGenerator.kt
@@ -23,6 +23,7 @@ package com.vaticle.dependencies.ide.rust
 
 import com.electronwill.nightconfig.core.Config
 import com.electronwill.nightconfig.toml.TomlWriter
+import com.vaticle.bazel.distribution.common.shell.Shell
 import com.vaticle.bazel.distribution.common.util.FileUtil.listFilesRecursively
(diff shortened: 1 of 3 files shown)


ide/rust/RepoCargoManifestGenerator.kt
```diff
@@ -95,11 +98,16 @@ class RepoCargoManifestGenerator(private val repository: File, private val bazel
 
     private inner class TargetCargoManifestGenerator(private val info: TargetSyncInfo) {
         fun generateManifest(): File {
+            if (info.containsGeneratedSources) buildTarget()
```
alexjpwalker, line 101: If the target has generated sources, build it to generate those sources so that `cargo build` can subsequently succeed.

This enables IDE tooling and debugging.

lolski (approved):

PR #401, opened 2022-12-05
Upgrade Kotlin, Compose Multiplatform

jamesreprise:

We want to upgrade to the latest versions of Compose Multiplatform and Kotlin for increased performance and functionality.

We've bumped `rules_kotlin`, which now provides its own method for importing compiler plugins.

We've also bumped Compose for improved stability and test reliability.

## Note
We still need to address the transitive dependencies issue that our fork of `rules_kotlin` currently patches out.


WORKSPACE
```diff
@@ -41,8 +41,14 @@ load("//library/maven:rules.bzl", "maven")
...
 kotlin_deps()
-load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories", "kt_register_toolchains")
-kotlin_repositories()
+load("@io_bazel_rules_kotlin//kotlin:repositories.bzl", "kotlin_repositories", "kotlinc_version")
+kotlin_repositories(
+    compiler_release = kotlinc_version(
+        release = "1.7.20",
+        sha256 = "5e3c8d0f965410ff12e90d6f8dc5df2fc09fd595a684d514616851ce7e94ae7d"
+    )
+)
```
jamesreprise, line 50: We specify our kotlinc version explicitly now, along with the SHA256 hash of the release.

WORKSPACE
```diff
@@ -41,8 +41,14 @@ load("//library/maven:rules.bzl", "maven")
...
+load("@io_bazel_rules_kotlin//kotlin:core.bzl", "kt_register_toolchains")
```
jamesreprise, line 51: Since we forked, `rules_kotlin` has adopted a more modular approach for functions rather than having everything be imported from (or aliased through) `kotlin:kotlin.bzl`.

builder/compose/deps.bzl
```diff
@@ -1,13 +1,6 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
 
 def deps():
-    http_file(
-        name = "org_jetbrains_compose_compiler",
-        urls = ["https://maven.pkg.jetbrains.space/public/p/compose/dev/org/jetbrains/compose/compiler/compiler-hosted/1.0.0-rc3/compiler-hosted-1.0.0-rc3.jar"],
-        sha256 = "1e628a81ad501bc737ca4ea193b8c5072a4baa022b3152aacb928bf076af8226",
-        downloaded_file_path = "compiler-hosted-1.0.0-rc3.jar",
-    )
-
```
jamesreprise, line 10: We don't need to specify the compose compiler this way anymore! We just specify it as a regular dependency, seen in //library/maven/artifacts.bzl.

builder/kotlin/deps.bzl
```diff
@@ -1,10 +1,11 @@
...
+    rules_kotlin_sha = "f033fa36f51073eae224f18428d9493966e67c27387728b6be2ebbdae43f140e"
+
     http_archive(
         name = "io_bazel_rules_kotlin",
-        urls = ["https://github.com/vaticle/rules_kotlin/archive/c2519b00299cff9df22267e8359784e9948dba67.zip"],
-        type = "zip",
-        strip_prefix = "rules_kotlin-c2519b00299cff9df22267e8359784e9948dba67",
-        sha256 = "1455f2ec4bf7ea12d2c90b0dfd6402553c3bb6cbc0271023e2e01ccdefb4a49a",
+        urls = ["https://github.com/bazelbuild/rules_kotlin/releases/download/v%s/rules_kotlin_release.tgz" % rules_kotlin_version],
+        sha256 = rules_kotlin_sha,
```
jamesreprise, line 10: We use the rules_kotlin from the bazelbuild team now rather than our own fork.

library/maven/artifacts.bzl
```diff
@@ -234,30 +234,31 @@ artifacts = {
...
-    "org.jetbrains.compose.ui:ui-text-desktop": "1.1.0-alpha04",
-    "org.jetbrains.compose.ui:ui-unit-desktop": "1.1.0-alpha04",
-    "org.jetbrains.compose.runtime:runtime-desktop": "1.1.0-alpha04",
-    "org.jetbrains.kotlin:kotlin-reflect": "1.6.0",
-    "org.jetbrains.kotlin:kotlin-stdlib-jdk7": "1.6.0",
-    "org.jetbrains.kotlin:kotlin-test": "1.6.0",
-    "org.jetbrains.kotlinx:kotlinx-coroutines-core": "1.6.0",
-    "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm": "1.6.0",
-    "org.jetbrains.kotlinx:kotlinx-coroutines-test": "1.6.0",
+    "org.jetbrains.compose.compiler:compiler": "1.3.2",
```
jamesreprise, line 237: Compose compiler as a maven import! Horray!

WORKSPACE
alexjpwalker: Interesting. Is this an approach favoured by `rules_kotlin` now?

In Rust, originally, we specified our `rustc` version explicitly, and then we changed that behaviour because each release of `rules_rust` has its own default `rustc` version number which doesn't change until we upgrade `rules_rust` again. Having said that, I'm not certain that was the right approach for us. On the one hand, it's good to be explicit about language versions; on the other hand, we (almost!) always just want the latest version, and in the latter case, it's arguably better for there to be a single source of truth for the language version - and that's whatever version is specified in `rules_rust` (or in our case here, `rules_kotlin`).

library/maven/artifacts.bzl
alexjpwalker: Great! Can we get a usage example?

alexjpwalker (changes requested):
Requested some changes

WORKSPACE
jamesreprise: I was mistaken having this here: we specify the version of kotlinc per WORKSPACE. dependencies should only be making the appropriate version of rules_kotlin available.

library/maven/artifacts.bzl
jamesreprise: No longer applicable in the same way, but I'll explain the new usage:
* We specify the compiler plugin as a dependency as above.
* We specify the compiler plugin as a `kt_compiler_plugin` for consumption by dependors in '//builder/compose/BUILD`.
* We make use of this plugin via methods such as `kt_jvm_library` adding the argument `plugins = ["@vaticle_dependencies//builder/compose:compiler_plugin"]`.

WORKSPACE
alexjpwalker: Hm. It makes sense that we can specify the `kotlinc` version per `WORKSPACE`; however, `dependencies` is also a self-contained repo that we are supposed to be able to run `bazel build //...` on and get a deterministic output. And it most certainly does use Kotlin binaries to build the project.

Given, therefore, that the build succeeds without a `kotlinc` version specified, I suppose we can assume that `rules_kotlin` behaves similarly to `rules_rust`, having a default `kotlinc` version that it uses?

Could we verify this?

library/maven/artifacts.bzl
alexjpwalker: Let's copy in the relevant code snippets (that we already wrote in the Studio PR) for reference to future readers here

WORKSPACE
jamesreprise: > Given, therefore, that the build succeeds without a kotlinc version specified, I suppose we can assume that rules_kotlin behaves similarly to rules_rust, having a default kotlinc version that it uses?
That's correct. When we call `kotlin_repositories` with no arguments, a default kotlinc version is chosen based on some hardcoded values in rules_kotlin's `//src/main/starlark/core/repositories:versions.bzl`.

library/maven/artifacts.bzl
jamesreprise: Done.

alexjpwalker (approved):

PR #407, opened 2023-01-12
Only include versioned labels in Maven artifacts.snapshot

### LINKED ISSUE #152: artifacts.snapshot produced by //library/maven:update contains duplicates
See the `artifacts.snapshot` file from `@graknlabs_verification` linked below.

https://github.com/graknlabs/verification/blob/82012a1904d838dd675b680985a25460353d8665/dependencies/maven/artifacts.snapshot

We can see that each dependency artifact is written twice: once with the version, and once without. Is this expected? This feels like a bug. Can we fix it so that it just outputs one artifact once just with the version?

alexjpwalker:

We now only include versioned labels in a Maven `artifacts.snapshot` generated by `//library/maven:update`.

Each non-versioned `@maven` target is a `jvm_import`. Each versioned `@maven` target is an `alias` that aliases the non-versioned target.

Therefore we can halve the size of `artifacts.snapshot` by changing the invocation
```
bazel query @maven//...
```
to
```
bazel query 'kind(alias, @maven//...)'
```

Closes #152 .

### DIFF
--- a/library/maven/update.kt
+++ b/library/maven/update.kt
@@ -7,7 +7,7 @@ import java.nio.file.Paths
 
 fun main() {
     val baseDir = Paths.get(System.getenv("BUILD_WORKSPACE_DIRECTORY"))
-    val snapshotCommand = listOf("bazel", "query", "@maven//...")
+    val snapshotCommand = listOf("bazel", "query", "kind(alias, @maven//...)")
     val snapshotFile = baseDir.resolve("dependencies").resolve("maven").resolve("artifacts.snapshot")
 
     println("-------------------------")


library/maven/update.kt
```diff
@@ -7,7 +7,7 @@ import java.nio.file.Paths
 
 fun main() {
     val baseDir = Paths.get(System.getenv("BUILD_WORKSPACE_DIRECTORY"))
-    val snapshotCommand = listOf("bazel", "query", "@maven//...")
+    val snapshotCommand = listOf("bazel", "query", "kind(alias, @maven//...)")
```
alexjpwalker, line 10: See PR description.

lolski (approved):

PR #411, opened 2023-01-17
Don't use Bazel in crate-universe lockfile update script

dmitrii-ubskii:

rules_rust's own way of repinning changed dependencies in `crates_repository`, i.e. by setting the `CARGO_BAZEL_REPIN` environment variable to `full` takes a prohibitively long time for minor changes to the cargo manifest. We instead opted for using `cargo generate-lockfile`, which takes an order of magnitude less time to complete (~10 s as opposed to 3-5 minutes).

Using the cargo executable provided by rules_rust seems to be the most correct approach to updating the lockfile. However, calling it directly requires Bazel to build the workspace. Since we have rust executables in dependencies (e.g. `//builder/grpc/rust:compile`), Bazel will attempt to fetch the crates. Fetching the crates will fail if the manifest and the lockfile don't match, which they won't, since we just changed the manifest and are attempting to regenerate the lockfile.

The options are:
1. use `CARGO_BAZEL_REPIN=full`, slow as it is;
2. use system `cargo`;
3. edit WORKSPACE temporarily from within `update.sh`;
4. download a `cargo` binary and use that to generate the lockfile.

We've chosen option 4 for this PR, as that is fast, doesn't rely on the user system as much as option 2, and isn't as hacky as option 3. It is, however, limited to darwin-x64 at present.

In `library/crates/update.sh`, fetch the cargo executable from repo.vaticle.com and regenerate the lockfile using that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ bazel-bin
 bazel-out
 bazel-dependencies
 bazel-testlogs
-cargo
+library/crates/cargo
 tool/bazelcache/.terraform/*
 target
 Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ bazel-bin
 bazel-out
 bazel-dependencies
 bazel-testlogs
+cargo
 tool/bazelcache/.terraform/*
 target
 Cargo.lock
--- a/library/crates/update.sh
+++ b/library/crates/update.sh
@@ -22,15 +22,11 @@ set -ex
 
 [[ $(readlink $0) ]] && path=$(readlink $0) || path=$0
 
-cargo_target=@rules_rust//rust/toolchain:current_cargo_files
-
-bazel build $cargo_target
-
-project_dir=$(bazel info workspace)
-cargo_relpath=$(bazel cquery $cargo_target --output starlark --starlark:expr="target.files.to_list()[0].path" 2>/dev/null)
-cargo=${project_dir}/${cargo_relpath}
-
 crates_home=$(cd "$(dirname "${path}")" && pwd -P)
 pushd "$crates_home" > /dev/null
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@ bazel-bin
 bazel-out
 bazel-dependencies
 bazel-testlogs
+cargo
```
alexjpwalker, line 8: This can be `library/crates/cargo` right?

library/crates/update.sh
```diff
@@ -22,15 +22,18 @@ set -ex
...
-cargo_relpath=$(bazel cquery $cargo_target --output starlark --starlark:expr="target.files.to_list()[0].path" 2>/dev/null)
-cargo=${project_dir}/${cargo_relpath}
-
 crates_home=$(cd "$(dirname "${path}")" && pwd -P)
 pushd "$crates_home" > /dev/null
-$cargo generate-lockfile
+if [ ! -x cargo ]; then
+    arch=$(bash --version | head -n1 | grep -o '\S\+$')  # (arch-vendor-os)
+    arch=${arch#(} && arch=${arch%)} # strip parentheses
+    if [[ $arch == x86_64-apple-darwin* ]]; then
```
alexjpwalker, line 30: Use `*-apple-darwin*`, otherwise this script will fail on an M1 Mac when in reality it should run using Rosetta.

alexjpwalker (changes requested):

library/crates/update.sh
dmitrii-ubskii: I'd rather be more explicit. Added a branch for `arm64-apple-darwin` instead

alexjpwalker (approved):

PR #413, opened 2023-01-18
Improve crate name detection in Rust IDE sync

alexjpwalker:

`ide/rust:sync` is now more reliable at correctly detecting the crate names of all dependencies. In particular, crates with hyphens (such as `async-trait`) should no longer cause issues.

Crate name detection has been reworked. Because `rules_rust` doesn't expose the original name of the crate downloaded from Cargo anywhere, we were unable to reliably get that crate name and put it into `Cargo.toml`. Now, we use string parsing against the target's label. This works because the label is (e.g):
```
@crates__async-trait-0.36.0//:async_trait
```
We take everything up to + excluding the first `.`, then take everything up to + excluding the last `-`, then strip `@crates__` from the front.

See also:
- https://github.com/vaticle/bazel-distribution/pull/367

### DIFF
--- a/distribution/deps.bzl
+++ b/distribution/deps.bzl
@@ -21,5 +21,5 @@ def vaticle_bazel_distribution():
     git_repository(
         name = "vaticle_bazel_distribution",
         remote = "https://github.com/vaticle/bazel-distribution",
-        commit = "cdf8f3e3e009b3690d0f467b1ecc5b4f122a74fa" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_bazel_distribution
+        commit = "16f85fa14818c61bd0ed58553ab5d63f2039985d" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_bazel_distribution
     )
--- a/ide/rust/sync_info_aspect.bzl
+++ b/ide/rust/sync_info_aspect.bzl
@@ -29,8 +29,11 @@ def _should_generate_dep_info(dependency):
 def _should_generate_build_dep_info(dependency):
     return dependency.kind in _TARGET_TYPES and _TARGET_TYPES[dependency.kind] == "build"
 
-def _is_universe_crate(rust_target):
-    return str(rust_target.label).startswith("@crates__")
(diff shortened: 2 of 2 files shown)


distribution/deps.bzl
```diff
@@ -21,5 +21,5 @@ def vaticle_bazel_distribution():
     git_repository(
         name = "vaticle_bazel_distribution",
         remote = "https://github.com/vaticle/bazel-distribution",
-        commit = "cdf8f3e3e009b3690d0f467b1ecc5b4f122a74fa" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_bazel_distribution
+        commit = "16f85fa14818c61bd0ed58553ab5d63f2039985d" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_bazel_distribution
```
alexjpwalker, line 24: We also update to the latest `bazel-distribution`, where a similar fix was just written for `assemble_crate`.

nikolaimerritt: I've taken a look at the Semantic Versioning specification (https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions). This should comply with any crate name where the name is proceeded by a valid semantic version.

ide/rust/sync_info_aspect.bzl
```diff
@@ -29,8 +29,11 @@ def _should_generate_dep_info(dependency):
 def _should_generate_build_dep_info(dependency):
     return dependency.kind in _TARGET_TYPES and _TARGET_TYPES[dependency.kind] == "build"
 
-def _is_universe_crate(rust_target):
-    return str(rust_target.label).startswith("@crates__")
+def _is_universe_crate(target):
+    return str(target.label).startswith("@crates__")
+
+def _universe_crate_name(target):
+    return str(target.label).split(".")[0].rsplit("-", 1)[0].removeprefix("@crates__")
```
nikolaimerritt, line 36: As per (https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions), this should comply with any crate name where the name is proceeded by a valid semantic version.

PR #421, opened 2023-02-13
Update checkstyle to use original authorship year

jamesreprise:

We've updated the checkstyle rules to use the style `2015-present` rather explicitly specifying the year and having to bump the year in the rest of the code every year.

We've made changes to `checkstyle_test`, adding a year parameter alongside the `license_type` parameter, specifying the year in which the code was authored. If no year is specified, it defaults to 2015.

### DIFF
--- a/tool/checkstyle/rules.bzl
+++ b/tool/checkstyle/rules.bzl
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2022 Vaticle
+# Copyright (C) 2018-present Vaticle
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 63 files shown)


tool/checkstyle/rules.bzl
```diff
@@ -20,32 +20,40 @@ def _checkstyle_test_impl(ctx):
     opts = ctx.attr.opts
     sopts = ctx.attr.string_opts
 
-    args = ""
-    inputs = []
+    license_file = "tool/checkstyle/config/checkstyle-file-%s.txt" % ctx.attr.license_type
 
-    license_file = "external/vaticle_dependencies/tool/checkstyle/config/checkstyle-file-%s.txt" % ctx.attr.license_type
+    if ctx.workspace_name != "vaticle_dependencies":
```
jamesreprise, line 25: We have to support checkstyles existing in vaticle_dependencies, in which case the license file won't be found in `external/vaticle_dependencies` (because this is vaticle_dependencies! The checkstyle was coming from inside the house!)

tool/checkstyle/rules.bzl
```diff
@@ -20,32 +20,40 @@ def _checkstyle_test_impl(ctx):
...
+        license_file = "external/vaticle_dependencies/" + license_file
+    expanded_license_file = ctx.actions.declare_file("checkstyle-file-expanded-%s.txt" % ctx.attr.license_type)
+    ctx.actions.run_shell(
+        outputs = [ expanded_license_file ],
+        inputs = ctx.files._license_files,
+        command = 'sed "s/YEAR/{}/g" {} > {}'.format(ctx.attr.year, license_file, expanded_license_file.path),
```
jamesreprise, line 31: We substitute in the year passed to us by the `year` argument.

tool/checkstyle/rules.bzl
jamesreprise: This hardcoded string check isn't ideal and seems a little fragile, but I haven't found a way to ask bazel if the rule implementation we're in is defined in the same workspace we're trying to check.

jamesreprise:
Based on an in-person conversation, we actually want to remove our use of the checkstyle tool altogether. The rationale is that it is a fairly heavyweight tool, and we're using it for something we could achieve with a simple regex. Additionally, we're moving towards languages that checkstyle doesn't support for static code analysis, its killer application.

In its place, we want a tool that regexes across a repo ensuring that all necessary files have the appropriate copyright header. A single rule should be sufficient to cover the entire repo.

We've aligned on the current string substitution approach with a stipulation that we change the value to be substituted from `YEAR` to `{YEAR}`.

PR #422, opened 2023-02-21
Generify IDE sync tool into Cargo sync tool

flyingsilverfin:

We refactor the previously explicitly bound IDE sync tool for converting bazel `rust_*` targets into Cargo manifests recursively, into a more generic Cargo sync tool.

* Move `//ide/rust:sync` into `//tool/cargo:sync`
* Refactor implementation of cargo manifest syncing to not reference the term "IDE"`
* Refactor implementation of cargo manifest syncing naming and structure:
  * Merge two kotlin files into one with nested classes. We now only have a `Syncer` which contains a `WorkspaceSyncer`, which in turn contains `ManifestGenerator`
  * We replace the term `Repository` with `Workspace`, since `Repository` is overloaded between Bazel and Git. A `Workspace` and bazel `Repository` is pretty much interchangable -- repositories are more general in that they are just code-containing directories that may not be bazel projects with their own WORKSPACE files
  * Replace the term `TargetSyncInfo` with `TargetProperties`, which produce/derive from files ending in `.cargo-sync.properties`

We also fix some bugs related to windows/unix interopability by splitting and generating strings containing newlines by the current system's newline character, instead of always using `\n` -- on windows we must use `\r\n`.


tool/cargo/Syncer.kt
```diff
@@ -0,0 +1,407 @@
...
+import java.nio.file.Path
+import java.util.*
+import java.util.concurrent.Callable
+import kotlin.io.path.Path
+import kotlin.system.exitProcess
+
+fun main(args: Array<String>): Unit = exitProcess(CommandLine(Syncer()).execute(*args))
+
+@CommandLine.Command(name = "sync", mixinStandardHelpOptions = true)
+class Syncer : Callable<Unit> {
```
flyingsilverfin, line 56: Entry point has not changed

tool/cargo/Syncer.kt
```diff
@@ -0,0 +1,407 @@
...
+
+    override fun call() {
+        logger = Logger(logLevel = if (verbose) DEBUG else ERROR)
+        shell = Shell(logger, verbose)
+        bazelOutputBase = Path(shell.execute(listOf(BAZEL, INFO, OUTPUT_BASE), workspaceDir).outputString().trim())
+
+        val rustTargets = rustTargets(shell, workspaceDir)
+        validateTargets(rustTargets)
+        loadRustToolchainAndExternalDeps(rustTargets)
+        vaticleRustWorkspaces().forEach { WorkspaceSyncer(it, logger, shell, bazelOutputBase).sync() }
```
flyingsilverfin, line 74: We now call a `WorkspaceSyncer`, previously called a `RepositorySyncer`

tool/cargo/Syncer.kt
```diff
@@ -0,0 +1,407 @@
...
+
+    private class WorkspaceSyncer(private val workspace: Path, private var logger: Logger, private var shell: Shell, var bazelOutputBase: Path) {
+
+        private val bazelBin = workspace.resolve(BAZEL_BIN).toRealPath().toFile()
+
+        fun sync() {
+            logger.debug { "Syncing $workspace" }
+            cleanupOldSyncProperties()
+            runSyncPropertiesAspect()
+            generateManifests()
```
flyingsilverfin, line 109: Now, instead of creating a new class here which handles the search of sync properties files, and then convertes each into manifests, the search of properties files is handled directly in this class. Each one is then converted to a Manifest with the `ManifestGenerator`

tool/cargo/Syncer.kt
```diff
@@ -0,0 +1,407 @@
...
+
+            companion object {
+                fun fromPropertiesFile(path: Path): TargetProperties {
+                    val props = Properties().apply { load(FileInputStream(path.toString())) }
+                    try {
+                        return TargetProperties(
+                                path = path,
+                                name = props.getProperty(Keys.NAME),
+                                label = props.getProperty(Keys.LABEL),
+                                type = Type.of(props.getProperty(Keys.TYPE)),
```
flyingsilverfin, line 331: note: removed some of the static imports that were not obvious to the the reader when new to the project -- this is preference though

tool/cargo/sync_properties_aspect.bzl
```diff
@@ -23,10 +23,10 @@ _TARGET_TYPES = {
     "_build_script_run": "build"
 }
 
-def _should_generate_dep_info(dependency):
+def _should_generate_dep_properties(dependency):
```
flyingsilverfin, line 26: All the methods have been renamed to replace `info` with `properties`

tool/cargo/Syncer.kt
```diff
@@ -0,0 +1,407 @@
...
+            const val CARGO_TOML = "Cargo.toml"
+            const val EXTERNAL = "external"
+            const val EXTERNAL_PLACEHOLDER = "{external}"
+            const val MANIFEST_SYNC_PROPERTIES_SUFFIX = ".cargo-sync.properties"
+        }
+
+        companion object {
+            const val GENERATED_FILE_NOTICE =
+"""
+# Generated by Vaticle Manifest sync tool.
```
alexjpwalker, line 388:
```suggestion
# Generated by Vaticle Cargo sync tool.
```

tool/cargo/Syncer.kt
alexjpwalker: My reasoning for statically importing `NAME`, `LABEL` etc. is that we always strive for canonical names (this is a principle discussed extensively in the Rust book, too!), and in this case it is quite clear that `props.getProperty(NAME)` is reading some kind of "name" property from the `Properties` object. `Keys` is redundant because, clearly, we're reading a key because that's what the method `getProperty` does.

tool/cargo/sync_properties_aspect.bzl
```diff
@@ -171,21 +170,21 @@ def _build_sync_info_file(target, ctx, source_files, crate_info):
...
 
-def _rust_ide_sync_info_aspect_impl(target, ctx):
+def _rust_cargo_sync_properties_aspect_impl(target, ctx):
     if ctx.rule.kind not in _TARGET_TYPES.keys():
         return struct(kind = ctx.rule.kind)
 
     sources = [f for src in getattr(ctx.rule.attr, "srcs", []) for f in src.files.to_list()]
     crate_info = _crate_info(ctx, target)
-    sync_info_file = _build_sync_info_file(target, ctx, sources, crate_info)
+    sync_info_file = _build_cargo_sync_properties_file(target, ctx, sources, crate_info)
```
alexjpwalker, line 179: Right. I'd (incorrectly!) said previously that the term "Info" carries no meaning in Bazel. This isn't quite accurate. [Providers](https://bazel.build/rules/lib/Provider) are conventionally named `.*Info`. [Structs](https://bazel.build/rules/lib/struct) are slightly less strongly-typed but often more lightweight and can be used in place of providers - as we do here, returning a `struct` from the `...aspect_impl` method as opposed to a provider.

That being said, this method was always a little awkward, because the struct was called "sync info", and the output file was called "sync info", and guess what, they contained different content. So now we can fix that situation.

But - now that we've resolved that ambiguity, let's make sure we apply it consistently! This variable, for instance, `sync_info_file`, should be renamed to `cargo_sync_properties_file`. (And maybe we should use `props` throughout for brevity?)

alexjpwalker (changes requested):

tool/cargo/Syncer.kt
flyingsilverfin: Alright, I've reverted this to be just using `NAME`, etc. directly :) Makes sense!

tool/cargo/sync_properties_aspect.bzl
flyingsilverfin: I just missed this variable seems like, have updating the naming here :)

alexjpwalker (approved):

tool/cargo/Syncer.kt
```diff
@@ -0,0 +1,419 @@
...
+        const val ASPECTS = "--aspects"
+        const val BAZEL = "bazel"
+        const val BAZEL_BIN = "bazel-bin"
+        const val BUILD = "build"
+        const val INFO = "info"
+        const val OUTPUT_BASE = "output_base"
+        const val OUTPUT_GROUPS_RUST_CARGO_SYNC_PROPERTIES = "--output_groups=rust-cargo-sync-properties"
+        const val QUERY = "query"
+        const val RUST_CARGO_SYNC_PROPERTIES_ASPECT = "@vaticle_dependencies//tool/cargo:sync_properties_aspect.bzl%rust_cargo_sync_properties_aspect"
+        const val RUST_TARGETS_DEPS_QUERY = "kind(rust_*, deps(kind(rust_*, //...)))" // deps of RUST_TARGETS_QUERY
```
flyingsilverfin, line 415: Note: we replaced `kind(rust_*, deps(//...))` with just searching the rust dependencies of the top-level rust dependencies, because calling `deps` seems to induce an unneeded build phase on Bazel (that happens to break compilation of RocksDB on Windows). In any case, this is not required, nor desirable in the top-level repository.

This could be simplified to: in each workspace, find all `rust_*` targets, and recursively find directly descendant `rust_*` targets. However, this implementation is more far more complex to implement. If we did implement it though, it works because of the idea that a Cargo project can only depend on Cargo projects recursively without breaks - we can't create a valid cargo project from something that looks like `rust_library -> genrule -> rust_library`.

PR #423, opened 2023-02-22
TypeDB Client Rust dependencies: async runtimes, tokio-stream

dmitrii-ubskii:

We add dependencies used by TypeDB Client Rust for async runtime independence tests.

* `tokio-stream`: utility, maps tokio mpsc channel receivers to streams;
* `http`: remove a layer of indirection;
* `smol` and `async-std`: popular async runtimes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ target
 Cargo.lock
 Cargo.toml
 !library/crates/Cargo.toml
+!library/crates/Cargo.lock
--- a/library/crates/Cargo.toml
+++ b/library/crates/Cargo.toml
@@ -25,6 +25,7 @@ path = "" # ignored by cargo generate-lockfile
 
 [dependencies]
 alcoholic_jwt = "=4091.0.0"
+async-std = { version = "=1.12.0", features = ["attributes"] }
 async-trait = "=0.1.59"
 axum = { version = "=0.5.15", features = ["ws"] }
 chrono = "=0.4.23"
@@ -34,6 +35,7 @@ cucumber = "=0.15.2"
 cxx = "=1.0.59"
 fluvio-helm = "=0.4.3"
 futures = { version = "=0.3.21", features = ["executor", "thread-pool"] }
+http = "=0.2.8"
 itertools = "=0.10.3"
 k8s-openapi = { version = "=0.16.0", features = ["v1_25"] }
 kube = { version = "=0.77.0", features = ["runtime", "derive"] }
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ target
 Cargo.lock
 Cargo.toml
 !library/crates/Cargo.toml
+!library/crates/Cargo.lock
```
dmitrii-ubskii, line 14: Crates-universe depends on this Cargo.lock to be present.

alexjpwalker: Why? How was it working before we had this?

library/crates/Cargo.toml
```diff
@@ -52,7 +54,9 @@ serde = "=1.0.152"
 serde_json = "=1.0"
 serial_test = "=0.9.0"
 sha2 = "=0.10.6"
+smol = "=1.3.0"
 tokio = { version = "=1.24.1", features = ["rt", "rt-multi-thread"] }
+tokio-stream = "=0.1.11"
```
alexjpwalker, line 59: Do we know why Tokio packages their stream library as a separate crate, as opposed to what they normally do, namely shipping it as a **feature** of `tokio` crate itself?

alexjpwalker (changes requested):
Also, please include a couple of the most important crate names in the PR title so that it's more "unique" (currently the same PR title could be applied to >10 different PRs).

Something like "TypeDB Client Rust dependencies (`async-std`, `tokio-stream`, etc.)"

library/crates/Cargo.toml
dmitrii-ubskii: TL;DR: tokio's minimum supported Rust version does not support streams.

https://docs.rs/tokio/latest/tokio/stream/

> Originally, we had planned to ship Tokio 1.0 with a stable Stream type but unfortunately the [RFC](https://github.com/rust-lang/rfcs/pull/2996) had not been merged in time for Stream to reach std on a stable compiler in time for the 1.0 release of Tokio. <...> [O]nce `Stream` has made it into the standard library and the MSRV period has passed, we will implement stream for our different types.

.gitignore
dmitrii-ubskii: `.gitignore` only affects files that aren't tracked.

alexjpwalker: So it was working because we'd generated `Cargo.lock` files on our local machines and that was allowing the update script to work?

dmitrii-ubskii: Because it was already present and tracked, modifications would be detected by git. 

I discovered that this was a problem when I removed Cargo.lock _through IDEA's project tree view,_ which runs `git rm ...`. `git` then promptly ignored the newly generated lockfile, because it was no longer tracked.

alexjpwalker (approved):

PR #428, opened 2023-03-16
Prepare ORTools for static compilation in Rust

flyingsilverfin:

We modify the OR-Tools dependency to be statically compiled into a project, rather than linked dynamically. This leverages the fact that OR-Tools is a Google repository that exposes Bazel targets to build with. As a result, we follow the model we use for RocksDB, which is to statically compile a platform-native library into our project on demand.

* Remove rules related to using dynmically linked OR-Tools libraries
* Include ORTools's github repository via Bazel, plus all of its  transitive dependencies
* Point the existing CXX bridge to generate the rust library based on the C++ sources instead of C++ header + pre-compiled dynamic libraries

### Justification

Like RocksDB, ORTools was initially considered as a dynamically linked library. However, we have discovered that the Rust+Bazel ecosystem does not support packaging dynamic libraries and binaries together elegantly, and is instead geared toward static compilation. 

For RocksDB, we rely on the `rocksdb-sys` crate to perform the compilation of RocksDB's C++ core into a Rust library. There is no such equivalent crate for ORTools, the existing crates are wrappers around dynamically linked ORTools distributions.

The solution for now is to leverage our existing work to build our own CXX bridge between Rust and C++. Instead of building this bridge on top of header files and dynamic libraries that come pre-compiled, we utilise ORTools' Bazel targets to statically compile the library for the current platform.

Note: as a consequence of this, projects using OR-tools from this repository will be much slower: compiling OR-tools can take some time, on top of the time required to clone its Github repository (~200mb). However, the compile time can be cut drastically by using a bazel build cache.


.bazelrc
```diff
@@ -1,3 +1,9 @@
+common --enable_platform_specific_config
```
flyingsilverfin, line 1: We set different build flags per-host. 
This is because OR-tools compilation requires C++20, and the flag to set this is different for windows MSVC compilers and non-MSVC compilers.

Note that this flag is another barrier to cross-compilation, if that ever comes back onto the table.

flyingsilverfin: This command basically is a note to bazel that it should should select platform-specific configurations according to the host platform

WORKSPACE
```diff
@@ -52,10 +52,10 @@ google_or_tools_mac()
...
 google_or_tools_windows()
 
-load("//library/ortools/cc:deps.bzl", "google_or_tools_mac", "google_or_tools_linux", "google_or_tools_windows")
-google_or_tools_mac()
-google_or_tools_linux()
-google_or_tools_windows()
+load("//library/ortools/cc:deps.bzl", "google_or_tools")
+google_or_tools()
+load("@com_google_protobuf//:protobuf_deps.bzl", google_or_tools_protobuf_deps = "protobuf_deps")
+google_or_tools_protobuf_deps()
```
flyingsilverfin, line 58: or tools has its own protobuf dependencies

library/ortools/cc/deps.bzl
```diff
@@ -16,51 +16,87 @@
...
-      "include/absl/**/*.h",
-      "include/absl/**/*.inc",
-      "include/google/protobuf/**/*.inc",
-      "include/google/protobuf/**/*.h",
-  ]),
-  strip_include_prefix = "include/",
-  visibility = ["//visibility:public"]
-)
-""".format(lib)
+def google_or_tools():
```
flyingsilverfin, line 21: `or_tools` and all of its dependencies. We get this from the template OR-tools bazel project: https://github.com/or-tools/bazel_or-tools

library/ortools/cc/deps.bzl
```diff
@@ -16,51 +16,87 @@
...
+hdrs = glob(['Eigen/**']),
+visibility = ['//visibility:public'],
+)
+"""
+    )
+
+    git_repository(
+        name = "com_google_ortools",
+        remote = "https://github.com/google/or-tools.git",
+        tag = "v9.3",
```
flyingsilverfin, line 101: we use a tag instead of `commit` + `shallow_since` to reduce the size of the initial git clone

flyingsilverfin:
Moving back to draft until I can come back to this to finish it up

PR #429, opened 2023-03-21
Config settings

sam-butcher:

We add `config_setting_group`s representing:
- Mac x86 64
- Mac ARM 64

We use the Skylib `config_setting_group` to combine the existing `config_setting` blocks for `is_mac` with `is_x86_64`, and `is_mac` with `is_arm64`


util/platform/BUILD
```diff
@@ -33,6 +35,22 @@ config_setting(
     ]
 )
 
+selects.config_setting_group(
```
lolski, line 38: It looks good. Shouldn't we place this near `is_mac`?

lolski (approved):

PR #431, opened 2023-03-31
Add Node 16 and PNPM to Vaticle Ubuntu Docker image; add `tower-http` crate lib

alexjpwalker:

We added Node 16 and PNPM to the Vaticle Ubuntu Docker image, and added the crate library `tower-http` with the `cors` feature enabled.

The new TypeDB Web server will be running Scully in the background through Node.js, which takes quite a while (a few minutes) to install and we don't want to repeat that process on each fresh build of the website.

It also uses CORS for local development so we need the `tower-http` crate dependency declared.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -26,5 +26,6 @@ build:
     deploy-ubuntu-docker:
       image: vaticle-ubuntu-22.04
       command: |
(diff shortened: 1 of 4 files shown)


alexjpwalker:
Reverting to draft as we now want to create a new image `vaticle/web` instead

PR #433, opened 2023-04-13
Bazel Cache Installation Tool

jamesreprise:

We've updated the Bazel cache installation tool, bringing it up to date with our latest approach.

We will run this file before every Factory job that wants to make use of the Bazel cache. This will not interfere with development or use by developers outside of Vaticle. Developers inside Vaticle will be able to make use of the cache by running this tool after exporting the appropriate environment variables. 

Repositories where this workflow will be common should add '.bazel-remote-cache.rc' and 'credential.json to .gitignore.

The tool now checks for the presence of non-empty $BAZEL_CACHE_CREDENTIAL and $BAZEL_CACHE_URL environment variables. If they are both found, we generate a 'bazel_cache_credential.json' file from the content of $BAZEL_CACHE_CREDENTIAL, which is expected to be a base64 encoded string of a JSON GCP credential file. If it does exist, but fails to deserialise, we exit with an error code 1.


tool/bazelinstall/rbe.sh
```diff
@@ -2,18 +2,20 @@
...
-    sudo mkdir -p $BAZEL_CACHE_CREDENTIAL && sudo chmod a+rwx $BAZEL_CACHE_CREDENTIAL
-    ( echo "build --remote_cache=$BAZEL_CACHE_URL"; echo "test --remote_cache=$BAZEL_CACHE_URL"; echo "run --remote_cache=$BAZEL_CACHE_URL" ) >> "$BAZEL_CACHE_CREDENTIAL/bazel-remote-cache.rc"
-    echo "The RBE credential has been installed!"
-    echo "Configuring Python..."
-    pyenv global system system
+    BAZEL_CACHE_CREDENTIAL_LOCATION=./credential.json
+    echo "A remote cache credential is found and will be saved to $BAZEL_CACHE_CREDENTIAL_LOCATION. Artifact and test results will be cached remotely."
+    BAZEL_CACHE_CREDENTIAL_JSON=$(echo "$BAZEL_CACHE_CREDENTIAL" | base64 -d $BAZEL_CACHE_CREDENTIAL_LOCATION)
+    if [[ -n "$BAZEL_CACHE_CREDENTIAL_JSON" ]]; then
+      ( echo "build --remote_cache=$BAZEL_CACHE_URL --google_credentials=$BAZEL_CACHE_CREDENTIAL_LOCATION"; echo "test --cache_test_results=no" ) >> ./.bazel-remote-cache.rc
```
jamesreprise, line 11: We don't want to cache test results by default as a result of this rc file having its build instructions inform the test instructions. See the option precedence instruction at: https://bazel.build/run/bazelrc

tool/bazelinstall/rbe.sh
```diff
@@ -2,18 +2,21 @@
...
+    BAZEL_CACHE_CREDENTIAL_LOCATION=./credential.json
+    echo "A remote cache credential is found and will be saved to $BAZEL_CACHE_CREDENTIAL_LOCATION. Artifact and test results will be cached remotely."
+    BAZEL_CACHE_CREDENTIAL_JSON=$(echo "$BAZEL_CACHE_CREDENTIAL" | base64 -d)
+    if [[ -n "$BAZEL_CACHE_CREDENTIAL_JSON" ]]; then
+      echo "$BAZEL_CACHE_CREDENTIAL_JSON" > $BAZEL_CACHE_CREDENTIAL_LOCATION
+      ( echo "build --remote_cache=$BAZEL_CACHE_URL --google_credentials=$BAZEL_CACHE_CREDENTIAL_LOCATION"; echo "test --cache_test_results=no" ) > ./.bazel-remote-cache.rc
+      echo "The remote cache has been installed!"
+    else
+      echo "The remote cache was not installed."
+      echo "\$BAZEL_CACHE_CREDENTIAL was non-empty, but failed to be decoded as base64. Confirm your value is valid."
```
jamesreprise, line 16: Would it be more in line with our engineering practices to give a non-zero exit code here? To get here, this script has to be called explicitly and the cache credential and cache URL environment variables have to be set, but the cache credential failed deserialisation from base64. 

As it currently stands, the user could think they've installed the cache but actually failed to. One would have to read their Factory/Bazel logs carefully to find out that this is the case.

jamesreprise:
Cache installation notes for circulation within Vaticle team:

## For each project:
### `automation.yml`
Add to the beginning of all job definitions:
`bazel run @vaticle_dependencies//tool/bazelinstall:rbe.sh`

### `.bazelrc`
Add before any other instructions:
`try-import ./.bazel-remote-cache.rc`

### `.gitignore`
```
# Bazel cache installation files
.bazel-remote-cache.rc
bazel_cache_credential.json
```

For pure build jobs:
We’d still like to perform builds end to end in jobs named ‘build’ or similar. We can switch the cache to write-only mode by passing the flag `--no_remote_accept_cached`.

Environment variables:
Set `BAZEL_CACHE_CREDENTIAL` to a base64 encoded version of your GCP service account key.
Set `BAZEL_CACHE_URL` to ` https://storage.googleapis.com/vaticle-engineers-bazel-remote-cache `

Set these within Factory and locally - we can also do this for other platforms such as CircleCI.

Quick Bazel command precedence intro:
`run` and `test` commands are made up of `build`. This means that any flags we apply to build will be present in `run` and `test`. This is, unless they are specifically overwritten in `run`, `test`, or in another definition of `build`. See more at https://bazel.build/run/bazelrc#bazelrc-syntax-semantics.

tool/bazelinstall/rbe.sh
jamesreprise: I've gone ahead and implemented this. Let me know if this is undesirable.

jamesreprise:
Note about GCP service account keys: I'll distribute these to everyone who needs them once we're happy with the approach and have merged this PR.

tool/bazelinstall/rbe.sh
```diff
@@ -2,18 +2,22 @@
...
-    ( echo "build --remote_cache=$BAZEL_CACHE_URL"; echo "test --remote_cache=$BAZEL_CACHE_URL"; echo "run --remote_cache=$BAZEL_CACHE_URL" ) >> "$BAZEL_CACHE_CREDENTIAL/bazel-remote-cache.rc"
-    echo "The RBE credential has been installed!"
-    echo "Configuring Python..."
-    pyenv global system system
+    BAZEL_CACHE_CREDENTIAL_LOCATION=./bazel_cache_credential.json
+    echo "A remote cache credential is found and will be saved to $BAZEL_CACHE_CREDENTIAL_LOCATION. Artifact and test results will be cached remotely."
+    BAZEL_CACHE_CREDENTIAL_JSON=$(echo "$BAZEL_CACHE_CREDENTIAL" | base64 -d)
+    if [[ -n "$BAZEL_CACHE_CREDENTIAL_JSON" ]]; then
+      echo "$BAZEL_CACHE_CREDENTIAL_JSON" > $BAZEL_CACHE_CREDENTIAL_LOCATION
+      ( echo "build --remote_cache=$BAZEL_CACHE_URL --google_credentials=$BAZEL_CACHE_CREDENTIAL_LOCATION"; echo "test --cache_test_results=no" ) > ./.bazel-remote-cache.rc
```
jamesreprise, line 12: We've disabled the caching of test results.

flyingsilverfin (approved):

PR #436, opened 2023-04-21
Use hyphens in bazel cache credential file name

jamesreprise:

We now use hyphens in the name of the Bazel cache credential file.

`bazel_cache_credential.json` -> `bazel-cache-credential.json`


flyingsilverfin (approved):

flyingsilverfin:
Not a problem but is there a motivation about why not using underscore?

jamesreprise:
> Not a problem but is there a motivation about why not using underscore?

Consistency with our house style. The files looked odd listed together before this change:
```
.bazel-remote-cache.rc
bazel_cache_credential.json
```

PR #453, opened 2023-07-10
SWIG rule fix: copy swig includes to bazel-bin

dmitrii-ubskii:

### DIFF
--- a/builder/swig/java.bzl
+++ b/builder/swig/java.bzl
@@ -45,8 +45,14 @@ def _swig_java_wrapper_impl(ctx):
         wrap_src = ctx.actions.declare_file("{}_wrap.c".format(module_name))
         swig_headers = []
 
+    includes = []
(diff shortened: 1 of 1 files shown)


builder/swig/java.bzl
```diff
@@ -45,8 +45,14 @@ def _swig_java_wrapper_impl(ctx):
         wrap_src = ctx.actions.declare_file("{}_wrap.c".format(module_name))
         swig_headers = []
 
+    includes = []
+    for f in ctx.files.includes:
+        include = ctx.actions.declare_file(f.short_path)
+        _copy_to_bin(ctx, f, include)
+        includes.append(include)
```
dmitrii-ubskii, line 52: Note: we could also copy none of the .i and .swgs to bazel-bin, but in that case SWIG cannot find the generated header instead.

PR #454, opened 2023-07-12
Upgrade bazel rules_* and GRPC dependencies

flyingsilverfin:

In order to use the latest Bazel and GRPC versions, we have to upgrade the various `rules_<lang>` dependencies during the upgrade of `grpc` dependencies. We also replace `stackb_rules_proto` (https://github.com/stackb/rules_proto) with `rules-proto-grpc` (https://rules-proto-grpc.com/en/latest/).

* Upgrade bazel-distribution to the latest, and match the `rules_<lang>` versions declared there
* Note: this allows using a hermitic python toolchain declaration in the WORKSPACE
* We upgrade `grpc`, and replace `stackb_rules_proto` with `rules-proto-grpc`. However, to make this setup work with the hermitic Python installation, we apply a patch outlined outlined from https://github.com/grpc/grpc/issues/31892. Without this patch, we can receive errors about missing python headers on some operating systems.
* We remove the installation of `bazel_toolchains` which were unused (?)


builder/grpc/grpc-hermitic-python.patch
flyingsilverfin, line 1: We have to patch the `com_github_grpc_grpc` dependency to allow using a hermitic Python installation, without which on some operating systems the build fails to find the required Python header files

WORKSPACE
```diff
@@ -33,10 +33,22 @@ rust_register_toolchains(edition = "2021", include_rustc_srcs = True)
 load("//builder/python:deps.bzl", python_deps = "deps")
 python_deps()
 
+load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
+py_repositories()
+python_register_toolchains(
+    name = "python39",
+    python_version = "3.9",
+)
```
flyingsilverfin, line 41: Install a hermitic python toolchain

builder/java/deps.bzl
```diff
@@ -1,9 +1,12 @@
...
+    RULES_JVM_EXTERNAL_SHA ="d31e369b854322ca5098ea12c69d7175ded971435e55c18dd9dd5f29cc5249ac"
+
     http_archive(
         name = "rules_jvm_external",
-        strip_prefix = "rules_jvm_external-3.2",
-        sha256 = "82262ff4223c5fda6fb7ff8bd63db8131b51b413d26eb49e3131037e79e324af",
-        url = "https://github.com/bazelbuild/rules_jvm_external/archive/refs/tags/3.2.zip",
+        strip_prefix = "rules_jvm_external-%s" % RULES_JVM_EXTERNAL_TAG,
+        sha256 = RULES_JVM_EXTERNAL_SHA,
+        url = "https://github.com/bazelbuild/rules_jvm_external/releases/download/%s/rules_jvm_external-%s.tar.gz" % (RULES_JVM_EXTERNAL_TAG, RULES_JVM_EXTERNAL_TAG)
```
flyingsilverfin, line 11: Upgrade `rules_jvm_external`

builder/nodejs/deps.bzl
```diff
@@ -1,10 +1,8 @@
...
 
 def deps(patch = []):
     http_archive(
         name = "build_bazel_rules_nodejs",
-        sha256 = "f2194102720e662dbf193546585d705e645314319554c6ce7e47d8b59f459e9c",
-        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/2.2.2/rules_nodejs-2.2.2.tar.gz"],
-        patches = patch,
-        patch_args = ["-p1"],
+        sha256 = "5dd1e5dea1322174c57d3ca7b899da381d516220793d0adef3ba03b9d23baa8e",
+        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/5.8.3/rules_nodejs-5.8.3.tar.gz"],
```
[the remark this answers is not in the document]
dmitrii-ubskii: https://github.com/bazelbuild/rules_nodejs/tree/5.8.3:
> As of [this announcement](https://blog.aspect.dev/rulesjs-launch), the maintainers of rules_nodejs have all moved to a faster, more compatible alternative: [rules_js](https://github.com/aspect-build/rules_js/).

Any reason why we're not moving straight to that?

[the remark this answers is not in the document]
dmitrii-ubskii: Note: cloud uses aspect rules_js https://github.com/vaticle/typedb-cloud/blob/development/WORKSPACE#L138

flyingsilverfin:
Closing due the following findings:

This upgrade required upgrading the `grpc` build rules to a much newer version. However, the newer version of the `stackb` rules were tricky to use and didn't appear to work very easily. Instead, this work proceeded with `rules-proto-grpc` and successfully got `typedb-protocol` ([PR](https://github.com/vaticle/typedb-protocol/pull/186)), `client-java` ([PR](https://github.com/vaticle/typedb-driver-java/pull/433)), `factory-tracing` ([PR](https://github.com/vaticle/factory-tracing/pull/58)), and `typedb` ([PR](https://github.com/vaticle/typedb/pull/6856)) to a working state, with the **exception of assembling the final distributions**.

The `java_grpc_library` rule provided by `rules-proto-grpc` are architected as macros that automatically pull in various dependencies via Bazel. See https://github.com/rules-proto-grpc/rules_proto_grpc/blob/7064b28a75b3feb014b20d3276e17498987a68e2/java/java_grpc_library.bzl#L26.

These dependencies are Bazel, not Maven dependencies. However, in order to build TypeDB's `zip` distributions, as well as deploy the Java client, we must provide maven coordinates for every java library we depend on. Because `rules-proto-grpc` does not depend on these repositories via Maven, our assembly rules fail.

One 'correct' solution could be to provide a mapping from Bazel targets to Maven coordinates to both the `java_deps` and the `assemble_maven` rules in [vaticle/bazel-distribution](github.com/vaticle/bazel-distribution). However, we consider reverting back to `stackb` simpler, since it produces source Java files instead of compiled JARs, which we can compile into our own java library and control its dependencies explicitly.

We also can gain some efficiency by delaying the Bazel and GRPC upgrade since we will drop Python support for compiling our protocol (after implementing native JNI-based drivers.)

PR #455, opened 2023-07-17
Update jvm_rules_external and Netty ssl dependencies

krishnangovindraj:
Update netty dependencies to be able to use OpenSSL.

* Updates `jvm_rules_external` to allow jar resolution based on platform
* Update netty dependencies to resolve a cyclic-dependency bug in the current version


builder/java/deps.bzl
```diff
@@ -1,9 +1,12 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_jar")
 
 def deps():
+    RULES_JVM_EXTERNAL_TAG = "5.3"
+    RULES_JVM_EXTERNAL_SHA ="d31e369b854322ca5098ea12c69d7175ded971435e55c18dd9dd5f29cc5249ac"
+
```
dmitrii-ubskii, line 6: This would arguably be a good project-wide pattern for us to adopt.

`TAG` => `VERSION` would be a more descriptive name here.

WORKSPACE
```diff
@@ -38,6 +38,12 @@ load("//builder/java:deps.bzl", java_deps = "deps")
 java_deps()
 load("//library/maven:rules.bzl", "maven")
 
+load("@rules_jvm_external//:repositories.bzl", "rules_jvm_external_deps")
+rules_jvm_external_deps()
+
+load("@rules_jvm_external//:setup.bzl", "rules_jvm_external_setup")
+rules_jvm_external_setup()
+
```
dmitrii-ubskii, line 46: Is this necessary in dependencies? Why didn't we have that here before?

krishnangovindraj: Sorry, this comment should have been part of the self-review.
It's the instructions here: https://github.com/bazelbuild/rules_jvm_external/releases/tag/5.3
And also synced up with what @flyingsilverfin is doing in his PR : https://github.com/vaticle/dependencies/pull/454

builder/java/deps.bzl
krishnangovindraj: Can do, this is following the suggestion from the Bazel repository here: https://github.com/bazelbuild/rules_jvm_external/releases/tag/5.3

WORKSPACE
krishnangovindraj: ^ Thank you, you're right. We don't need it in dependencies.
I've removed it now.

dmitrii-ubskii (approved):

PR #447, opened 2023-06-14
Cbindgen swig

dmitrii-ubskii:
Please replace every line in curly brackets { like this } with an appropriate description, and remove this line.

## What is the goal of this PR?

{ Please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. }

## What are the changes implemented in this PR?

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above, and reference the GitHub issues to be automatically closed, such like 'closes #number'. }

PR #460, opened 2023-08-10
Introduce SWIG Python rules

dmikhalin:

We introduce Python wrapper rules for SWIG in almost the same way as it has been done for Java in #447.

We also updated the `rules_python` version to 0.24.0 and replaced deprecated `pip_install()` with `pip_parse()`.


builder/swig/python.bzl
```diff
@@ -0,0 +1,172 @@
...
+            default = Label("@swig//:swig"),
+            allow_single_file = True,
+            executable = True,
+            cfg = "host",
+        ),
+    },
+)
+
+
+def swig_python_wrapper(**kwargs):
```
flyingsilverfin, line 126: Where are we passing/using a `select` in `_swig_python_wrapper`?

builder/swig/python.bzl
```diff
@@ -0,0 +1,172 @@
...
+    def swig_cc_binary(shared_lib_filename):
+        # name doesn't accept select() either
+        native.cc_binary(
+            name = shared_lib_filename,
+            deps = [lib, swig_wrapper_name],
+            srcs = [swig_wrapper_name],
+            linkshared = True,
+        )
+
+    swig_cc_binary_so = swig_cc_binary(shared_lib_name + ".so")
```
flyingsilverfin, line 156: By having it outside of a `select`, will we always have a `.so` rule instantiated? Even if we only assign it to a variable? I'm not sure what happens in this case. If we always end up with a `.so` rule declared then that's not correct.

dmikhalin: If we have this declaration inside `select` (twice), we get an error `cc_binary rule 'native_client_python.so' in package 'rust' conflicts with existing cc_binary rule`. It means that all rules in `select` are created (we have checked it). So the `select` block can be replaced with a simple rules creation.

builder/swig/python.bzl
dmikhalin: Yes, we don't use `select` here, it is an obsolete comment from java.

flyingsilverfin (approved):

PR #464, opened 2023-08-31
Allow linking SWIG Python for different platforms

dmikhalin:

We updated some dependencies to correspond with a newer `rules-rust` version, added a linux-arm64 platform, and fixed linking errors on Windows and errors with incorrect `libpython` dependency in the generated dynamic library.

### DIFF
--- a/builder/swig/python.bzl
+++ b/builder/swig/python.bzl
@@ -153,19 +153,40 @@ def swig_python(name, lib, shared_lib_name=None, linkopts=(), **kwargs):
             linkopts = linkopts,
         )
 
-    # FIXME: It should be possible to remove extensions and select
-#    swig_cc_binary(shared_lib_name + ".so")
-#    swig_cc_binary(shared_lib_name + ".dll")
-
     swig_cc_binary(shared_lib_name)
-
-#    native.alias(
-#        name = shared_lib_name,
-#        actual = select({
-#            "@vaticle_dependencies//util/platform:is_mac": (shared_lib_name + ".so"),
-#            "@vaticle_dependencies//util/platform:is_linux": (shared_lib_name + ".so"),
-#            "@vaticle_dependencies//util/platform:is_windows": (shared_lib_name + ".dll"),
-#        })
-#    )
-
     native.py_library(name = name, srcs = [swig_wrapper_name], data = [shared_lib_name])
+
+
+def _dyn_lib_impl(ctx):
+    output_file = ctx.actions.declare_file(ctx.attr.out)
+    _copy_to_bin(ctx, ctx.files.src[0], output_file)
+    return [
+        DefaultInfo(files = depset([output_file])),
+    ]
+
+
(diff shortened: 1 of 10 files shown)


builder/swig/python.bzl
```diff
@@ -24,7 +24,8 @@ def _copy_to_bin(ctx, src, dst):
 
 def _swig_python_wrapper_impl(ctx):
     module_name = getattr(ctx.attr, "class_name", ctx.attr.name)
```
flyingsilverfin: Probably should be `module_name`

builder/swig/python.bzl
```diff
@@ -133,28 +148,63 @@ def swig_python(name, lib, shared_lib_name=None, **kwargs):
...
+            mandatory = True,
+        ),
+        "src": attr.label(
+            mandatory = True,
+        ),
+    }
+)
+
+
+def py_native_lib_rename(name, out, src, visibility, **kwargs):
```
flyingsilverfin, line 196: Let's put a comment in the code here to save ourselves lots of pain in the future

builder/swig/python.bzl
```diff
@@ -123,38 +134,73 @@ swig_python_wrapper = rule(
...
-        actual = select({
-            "@vaticle_dependencies//util/platform:is_mac": (shared_lib_name + ".so"),
-            "@vaticle_dependencies//util/platform:is_linux": (shared_lib_name + ".so"),
-            "@vaticle_dependencies//util/platform:is_windows": (shared_lib_name + ".lib"),
-        })
+        deps = [lib, swig_wrapper_name],
+        srcs = [swig_wrapper_name],
+        linkshared = True,
+        linkopts = select({
+            "@vaticle_dependencies//util/platform:is_windows": ["ntdll.lib"],
```
flyingsilverfin, line 162: Let's leave a comment here in the code

flyingsilverfin (approved):

flyingsilverfin:
Title / goal should mention SWIG and Python to provide context

builder/swig/python.bzl
dmikhalin: `ntdll.lib` is necessary when linking in Windows, so we need to specify this `linkopts` argument for the linker.

builder/swig/python.bzl
dmikhalin: The generated dynamic library has to be copied into the output directory in order to be accessible by other rules.
We also choose the correct extension for the file.

builder/swig/python.bzl
```diff
@@ -123,38 +134,73 @@ swig_python_wrapper = rule(
...
+            "//conditions:default": [],
+        }),
+        copts = select({
+            "@vaticle_dependencies//util/platform:is_mac": ["-undefined", "dynamic_lookup"],
```
dmikhalin, line 166: We need to allow weak import when linking on MacOS.

flyingsilverfin (approved):

PR #475, opened 2023-09-28
Java native artifact rules update to reflect CPU architecture

dmitrii-ubskii:

We update native java artifact rules to allow for all five currently supported platforms: mac-arm64 and -x64, linux-arm64 and -x64, and windows-x64.


builder/compose/deps.bzl
dmitrii-ubskii, line 1: JDK16 is no longer available to download.

util/platform/BUILD
dmitrii-ubskii, line 1: Moved to bazel-distribution

flyingsilverfin (approved):

PR #479, opened 2023-10-10
Introduce dependencies and rules for C and C++

krishnangovindraj:
Add google's GTEST as dependency for use in our C and C++ code in typedb-driver.

* Introduces google's gtest library as a dependency.

### DIFF
--- a/builder/cpp/deps.bzl
+++ b/builder/cpp/deps.bzl
@@ -3,7 +3,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 def deps():
     http_archive(
         name = "gtest",
-        sha256 = "1f357c27ca988c3f7c6b4bf68a9395005ac6761f034046e9dde0896e3aba00e4",
-        strip_prefix = "googletest-1.14.0",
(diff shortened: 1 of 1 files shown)


library/cpp/clang_format.sh
dmitrii-ubskii, line 1: This should move to dependencies (or bazel-distribution) later

dmitrii-ubskii: Also, I think the way this is set up, this should run like maven update, i.e. format all the C files. In CI, we should have a job that runs this and fails on a non-empty git diff.

krishnangovindraj: (We discussed this in standup, but copying here)
This now runs more like a check style test, with the user running clang-format from their IDE.
Will add it to CI the same way we do checkstyle:
`bazel test $(bazel query 'kind(clang_format_test, //...)')`

krishnangovindraj: Why `bazel-distribution`
> Bazel rules for assembling and deploying software distributions 

Whereas the checkstyle rule & maven update rules both live here in dependencies.

dmitrii-ubskii (approved):

PR #489, opened 2023-12-18
Move typedb_java_test rule to dependencies

dmitrii-ubskii:

Move TypeDB test helper rules from typedb-common to dependencies.

The `maven()` rule receives an additional `internal_artifacts` argument so we can depend on internal maven dependencies without pinning them additionally in this repository.

### DIFF
--- a/builder/java/rules.bzl
+++ b/builder/java/rules.bzl
@@ -140,7 +140,7 @@ def typedb_java_test(name, server_artifacts, console_artifacts = {},
     native.java_test(
         name = name,
         deps = depset(deps).to_list() + native_dependencies,
-        classpath_resources = depset(classpath_resources).to_list(),
+        classpath_resources = depset(classpath_resources + ["@vaticle_dependencies//builder/java:logback"]).to_list(),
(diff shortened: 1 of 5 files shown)


builder/java/rules.bzl
```diff
@@ -111,3 +130,62 @@ def native_dep_for_host_platform(name):
...
+
+def typedb_java_test(name, server_artifacts, console_artifacts = {},
+                      native_libraries_deps = [], deps = [], classpath_resources = [], data = [], args = [], **kwargs):
+    native_server_artifact_paths, native_server_artifact_labels = native_artifact_paths_and_labels(server_artifacts)
+    native_console_artifact_paths, native_console_artifact_labels = native_artifact_paths_and_labels(console_artifacts, mandatory = False)
+    native_dependencies = get_native_dependencies(native_libraries_deps)
+
+    native.java_test(
+        name = name,
+        deps = deps + native_dependencies,
```
dmitrii-ubskii, line 142: Note: we no longer implicitly depend on the runner library, as there is no longer a single runner library.

flyingsilverfin (approved):

PR #490, opened 2023-12-19
Add doxygen rules for cpp driver docs generation

krishnangovindraj:
Adds a bazel rule to generate docs for C++ code using doxygen 

Introduces the rule `doxygen_docs` rule in `//builder/cpp:rules.bzl` 
* Expects doxygen to be present on the system
* Uses a doxyfile template as configuration, And a python script to replace the relevant config lines.


krishnangovindraj:
Moving these into driver.

PR #502, opened 2024-02-01
Stripe Java and Ktor JSON dependencies

sam-butcher:

We add dependencies for performing JSON serialization during Ktor HTTP requests and contacting Stripe with Kotlin.

We add the required libraries: `stripe-java`, `ktor-client-content-negotiation-jvm`, and `ktor-serialization-gson-jvm`


iliavolkovspb (approved):
Looks good to me

PR #520, opened 2024-04-02
Add support for multiple names of version file in version bump tool

jamesreprise:

Our version bump tool now supports two types of file - text and package.json. This tool will bump the relevant version of the file and write it to disk.

We've added support for text and package.json. We've removed the git commit and push functionality.

## Notes
Since no one is using this tool, we have not reintroduced the commit and push functionality elsewhere, but doing so should be trivial. We found that having the incrementing, writing, committing and pushing happening within a single tool was too much complexity in a single tool and ended up hampering flexibility of use. We found we wanted to bump multiple files using this single tool but not produce multiple commits (and multiple pushes).


package.json
iliavolkovspb, line 1: Do we need this file in this repository?

PR #526, opened 2024-04-04
Update reference to a fixed bazel distr with green build

farost:
We update the reference to the bazel-distribution with a [commit](https://github.com/vaticle/bazel-distribution/commit/425493c9a92ef894b99b3b8071886dd2af8098cf), that fixes Bazel build issues.


krishnangovindraj (approved):
lgtm

PR #547, opened 2024-06-12
Add a universal aspect_bazel_lib dependency for csharp and nodejs rules

farost:

`typedb-driver` and `typedb-protocol` builds have been broken after a recent `ts` rules [update](https://github.com/vaticle/dependencies/commit/73f3514) because of the updated `aspect_bazel_lib` dependency. 

To fix the current issue and prevent its reproduction, we add a universal `aspect_bazel_lib` dependency with a specific version, which is maximal for all the dependent rules in `dependencies`. This dependency should be explicitly declared before using any rule that depends on `aspect_bazel_lib` of any version and updated every time a rule uses a newer version of it as its dependency. 

We add a universal `aspect_bazel_lib` macro with a single `http_archive` call for this dependency. We also add invocations of this macro for all the rules' dependencies that depend on `aspect_bazel_lib`. It lets us not worry about the relative order of these rules declarations in `WORKSPACE`s as only the `aspect_bazel_lib()`'s version will be used for all these rules.

The fix works because Bazel uses the first declaration of a dependency. This way, if we have multiple `http_archive`s of `aspect_bazel_lib`, only the first one is used for every rule depending on it. 

It can be a problem if this common dependency becomes not backward-compatible and we cannot make all of our dependent rules use a single version. However, everything is fine while it works. 

A similar approach can be used for any other common dependency if we face such issues in the future.

[An example of a fixed build for protocol](https://factory.vaticle.com/farost/typedb-protocol/44ae6769893f8eed77e0906c141166012381b179)


builder/aspect/deps.bzl
```diff
@@ -0,0 +1,21 @@
...
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# This macro ensures a consistent version of the aspect_bazel_lib dependency across multiple deps (csharp, nodejs) used in this repo.
+#
+# To prevent conflicts arising from different versions of aspect_bazel_lib expected by these rules,
+# always call this macro before loading any of the mentioned rules' dependencies. This guarantees that the version
+# specified here is used universally.
+#
+# Ideally, the version of aspect_bazel_lib should be the maximum among the versions required by its dependents.
+# This approach assumes that aspect_bazel_lib maintains backward compatibility.
```
dmitrii-ubskii, line 14: Is this comment still accurate? This rule should not be called explicitly in the `WORKSPACE` file.

dmitrii-ubskii (approved):

builder/aspect/deps.bzl
farost: I believe it is. I mention `call this macro before loading any of the mentioned rules' dependencies` which can be both `WORKSPACE` and other macros, right? We just use it inside the `dependencies` macros. It is still needed to be propagated if we add new rules that depend on `aspect_bazel_lib` and updated if it's updated in any dependent rule.

PR #553, opened 2024-08-09
Add paste crate to dependencies

farost:

We add `paste` crate to our Cargo dependencies: https://crates.io/crates/paste. This crate allows us to get more abilities in Rust code generation. 

[Example of usage in TypeDB 3.0](https://github.com/farost/typedb/blob/3.0-thing-commit-time-validation/concept/type_/type_manager/validation/operation_time_validation.rs#L281), where we need to generate multiple intermediate functions inside a single macro to reduce copy-paste and additional macro calls.

### DIFF
--- a/library/crates/Cargo.lock
+++ b/library/crates/Cargo.lock
@@ -81,9 +81,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.14"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "418c75fa768af9c03be99d17643f93f79bbba589895012a80e3452a19ddda15b"
+checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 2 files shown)


dmitrii-ubskii (approved):

library/crates/Cargo.lock
farost, line 1: I've used the most recent version of `paste`, which updated other dependencies we have, and my local build for 3.0 has not been broken. I haven't tested it with other repos using Cargo, though (and might do something wrong as I hadn't updated any cargo dependencies before). We can wait until sync-dependencies to see or you can point me to specific repos to test with the updated lock file.

farost: Nvm, I get the same dependency updates if I run `./library/crates/update.sh` in `development`. 

That's the diff of this command after I add the new crate:
```diff
diff --git a/library/crates/Cargo.lock b/library/crates/Cargo.lock
index 7e17ac3..78a2d27 100644
--- a/library/crates/Cargo.lock
+++ b/library/crates/Cargo.lock
@@ -4762,6 +4762,7 @@ dependencies = [
  "maybe-async",
  "mime",
  "mime_guess",
+ "paste",
  "pest",
  "pest_derive",
  "pprof",
diff --git a/library/crates/Cargo.toml b/library/crates/Cargo.toml
index c2517aa..fa71347 100644
--- a/library/crates/Cargo.toml
+++ b/library/crates/Cargo.toml
@@ -39,6 +39,7 @@ macro_rules_attribute = "0.2.0"
 maybe-async = "0.2.7"
 mime = "0.3.16"
 mime_guess = "2.0.4"
+paste = "1.0.15"
 pest = "2.7.4"
 pest_derive = "2.7.4"
 pprof = { version = "0.13.0", features = ["flamegraph", "criterion"] }

```

PR #570, opened 2025-03-04
Update dependencies for 3.1.0

farost:
## Usage and product changes
Update dependencies prior to release 3.1.0. In details:
* Update dependencies without pinned versions.
* Update the `rocksdb` crate
* Pin the version of the `cc` create to avoid breaking `rocksdb` Bazel build.

## Motivation
Make sure that the dependencies are in a stable and non-conflicting state

## Implementation
See comments.

### DIFF
--- a/library/crates/Cargo.toml
+++ b/library/crates/Cargo.toml
@@ -54,7 +54,7 @@ prost-build = "^0.13"
 pwhash = "1.0.0"
 rand = "0.8.5"
 regex = "1.6.0"
-rocksdb = { version = "0.22.0", features = ["lz4"], default_features = false }
+rocksdb = { version = "0.23.0", features = ["lz4"], default_features = false }
 rpassword = "7.3.1"
 rustyline= { version = "15.0.0", features = ["with-file-history", "custom-bindings"], default_features = false }
 same-file = "1.0.6"
--- a/library/crates/Cargo.toml
+++ b/library/crates/Cargo.toml
@@ -54,7 +54,7 @@ prost-build = "^0.13"
 pwhash = "1.0.0"
 rand = "0.8.5"
 regex = "1.6.0"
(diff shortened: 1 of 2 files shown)


library/crates/Cargo.toml
```diff
@@ -54,7 +54,7 @@ prost-build = "^0.13"
 pwhash = "1.0.0"
 rand = "0.8.5"
 regex = "1.6.0"
-rocksdb = { version = "0.22.0", features = ["lz4"], default_features = false }
+rocksdb = { version = "0.23.0", features = ["lz4", "bindgen-runtime"], default_features = false }
```
farost, line 57: Rocksdb 0.23.0 introduced `bindgen-runtime` and `bindgen-static`. `bindgen-runtime` used to be the only and the default option, and `bindgen-static` did not exist. Now, by using `default_features`, we turn off `bindgen-runtime`, and we don't have `bindgen` at all... Weird.

farost: It's not necessary to update it now, but since we discovered and solved this upgrade issue, I think it's valuable to grab it.

library/crates/Cargo.toml
```diff
@@ -16,6 +16,7 @@ alcoholic_jwt = "4091.0.0"
 async-stripe = { version = "0.39.1", features = ["runtime-tokio-hyper"] }
 async-trait = "0.1.59"
 axum = { version = "0.5.15", features = ["ws"] }
+cc = "=1.2.15"
```
farost, line 19: We pin the `cc` version. `cc@1.2.16` changes something so librocksdb_sys starts having issues with build through Bazel. It was discussed in Discord: https://discord.com/channels/665254494820368395/672580553677078548/1346804418489618443
There, we decided that pinning `cc` does not introduce actual issues for us (hopefully, it won't in the near future), so it's a valid option of going forward.

flyingsilverfin (approved):

PR #573, opened 2025-04-02
Add declared crate features to the Rust sync tool

farost:
## Usage and product changes
Introduce `declared_features` to the Rust sync tool, parsed from the tags of Bazel `rust_` targets, to generate `Cargo.toml` files with predeclared features that are not automatically used in the dependent targets (this happens with `crate_features`).

## Motivation
We need a way to declare features without their automatic usage to be able to activate them manually through `cargo` only when needed.

## Implementation
Parse `tags` in order to search for a new tag containing a list of declared features. Save these features into a new separate field which is then passed to the crate features declaration.

### DIFF
--- a/builder/rust/cargo/project_aspect.bzl
+++ b/builder/rust/cargo/project_aspect.bzl
@@ -19,15 +19,16 @@ _BIN_ENTRY_POINT = "main.rs"
 _LIB_ENTRY_POINT = "lib.rs"
 
 CrateInfo = provider(fields = [
-    "kind",            # str
-    "crate_name",      # str
-    "workspace_name",  # str
-    "crate_path",      # str
-    "version",         # str
-    "features",        # List[str]
-    "deps",            # List[target]
-    "transitive_deps", # List[target]
-    "build_deps",      # List[target]
+    "kind",              # str
+    "crate_name",        # str
+    "workspace_name",    # str
+    "crate_path",        # str
+    "version",           # str
+    "features",          # List[str]
+    "declared_features", # List[str]
+    "deps",              # List[target]
+    "transitive_deps",   # List[target]
+    "build_deps",        # List[target]
 ])
 
 CargoProjectInfo = provider(fields = [
@@ -102,6 +103,7 @@ rust_cargo_properties_aspect = aspect(
 )
 
 def _crate_info(ctx, target):
+    declared_features = []
     if _is_universe_crate(target):
         crate_name = str(target.label).split(".")[0].rsplit("-", 1)[0].removeprefix("@crates__")
     else:
@@ -109,6 +111,9 @@ def _crate_info(ctx, target):
         for tag in ctx.rule.attr.tags:
             if tag.startswith("crate-name"):
                 crate_name = tag.split("=")[1]
+            elif tag.startswith("declared-features"):
+                feature_str = tag.split("=")[1]
+                declared_features = [f.strip() for f in feature_str.split(",") if f.strip()]
 
     workspace_name = target.label.workspace_name
     crate_path = target.label.package
@@ -122,6 +127,7 @@ def _crate_info(ctx, target):
         crate_path = crate_path,
         version = getattr(ctx.rule.attr, "version", "0.0.0"),
         features = getattr(ctx.rule.attr, "crate_features", []),
+        declared_features = declared_features,
         deps = deps,
         transitive_deps = transitive_deps,
         build_deps = _crate_build_deps(ctx, target),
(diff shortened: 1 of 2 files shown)


builder/rust/cargo/project_aspect.bzl
```diff
@@ -122,6 +127,7 @@ def _crate_info(ctx, target):
         crate_path = crate_path,
         version = getattr(ctx.rule.attr, "version", "0.0.0"),
         features = getattr(ctx.rule.attr, "crate_features", []),
+        declared_features = declared_features,
```
farost, line 130: If I add them to `features`, it will also be propagated to the dependants.

builder/rust/cargo/project_aspect.bzl
```diff
@@ -19,15 +19,16 @@ _BIN_ENTRY_POINT = "main.rs"
...
-    "deps",            # List[target]
-    "transitive_deps", # List[target]
-    "build_deps",      # List[target]
+    "kind",              # str
+    "crate_name",        # str
+    "workspace_name",    # str
+    "crate_path",        # str
+    "version",           # str
+    "features",          # List[str]
+    "declared_features", # List[str]
```
dmitrii-ubskii, line 28: Maybe rename `features` => `enabled_features` and `declared_features` => `features`?

builder/rust/cargo/project_aspect.bzl
```diff
@@ -228,6 +234,7 @@ def _get_properties(target, ctx, source_files, crate_info):
     properties["name"] = crate_info.crate_name
     properties["path"] = crate_info.crate_path
     properties["features"] = ",".join(crate_info.features)
+    properties["declared_features"] = ",".join(crate_info.declared_features)
```
dmitrii-ubskii, line 237: The key should have a `.` instead of the `_` to conform to the established naming convention.

tool/ide/RustManifestSyncer.kt
```diff
@@ -650,6 +658,7 @@ class RustManifestSyncer : Callable<Unit> {
                 const val EDITION = "edition"
                 const val ENTRY_POINT_PATH = "entry.point.path"
                 const val FEATURES = "features"
+                const val DECLARED_FEATURES = "declared_features"
```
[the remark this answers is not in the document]
dmitrii-ubskii: `target.name`: bazel target name, e.g. `//server:server` -> `server`
`path`: path to the crate root in the bazel workspace, e.g. `ir` or `tests/behaviour/steps`
`entry.point.path`:  usually `lib.rs` or `main.rs`, relative to `path`

`workspace_name` (not a key in `.cargo.properties`, but rather a part of the dependency spec): from bazel workspace, e.g. `@typeql` -> `typeql`; not relevant for crates.io dependencies or internal dependencies, only used for cross-repo depednencies like typeql.

dmitrii-ubskii (changes requested):

builder/rust/cargo/project_aspect.bzl
farost: I'm not sure... It might be confusing for a single crate, because we still don't put a default value for it even if it's `enabled`.

builder/rust/cargo/project_aspect.bzl
farost: Sure, I was confused by the naming below

builder/rust/cargo/project_aspect.bzl
farost: What if this tag is declared on the final binary?
```
@@ -42,6 +42,9 @@ rust_binary(
         "@crates//:sentry",
         "@crates//:tokio",
     ],
+   crate_features = ["itsnotenabled"],
 )

```
*sync*
```
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,9 @@
 # Generated by TypeDB Cargo sync tool.
 # Do not modify this file.
 
-features = {}
+
+[features]
+       itsnotenabled = []
 
```

dmitrii-ubskii: Not sure what the question is. `itsnotenabled` is indeed not enabled. We don't have a way to specify which features are enabled by default for a crate.

farost: That's why I find the name `enabled_features` confusing in `project_aspect.bzl`

farost: And that's why I find the name `enabled_features` confusing in this context, too... But I don't mind too much

farost: Done

dmitrii-ubskii (approved):

PR #579, opened 2025-06-23
Add prost-types and update cargo dependencies

farost:
## Usage and product changes
Add [prost-types](https://crates.io/crates/prost-types) to Cargo dependencies.

## Motivation
This crate allows us to manually encode and decode protobuf (`prost`) messages. This is useful for less type-safe encoding of messages, e.g., to put more structured details into protobuf error statuses, which will be used in TypeDB Cluster.


farost:
Not needed yet

PR #603, opened 2026-04-15
Keep track of library types in cargo sync tool

dmitrii-ubskii:
## Usage and product changes

We keep track of the specific `crate-type` based on the rules_rust target name (e.g. `rust_shared_library => cdylib`) and emit them during the manifest sync.

We also fix cyclic dependency between targets in a package, and update the Rust version to 1.93.


builder/rust/cargo/project_aspect.bzl
```diff
@@ -13,7 +13,17 @@ _TARGET_TYPES = {
...
-_DEFAULT_RUST_EDITION = "2021"
+_CRATE_TYPES = {
+    "rust_library": "lib",
+    "rust_shared_library": "cdylib",
+    "rust_static_library": "staticlib",
+    "rust_binary": "bin",
+    "rust_proc_macro": "proc-macro",
+    "rust_test": "test",
+    "_build_script_run": "N/A"
+}
```
dmitrii-ubskii, line 24: A lot easier to handle when it's separate from the target types which don't care what kind of library a library is.

tool/ide/RustManifestSyncer.kt
```diff
@@ -220,9 +221,13 @@ class RustManifestSyncer : Callable<Unit> {
...
-                val parent = nonTestProperties.values.filter { it.path.parentFile.toPath().equals(path.parent) };
+                var parent = nonTestProperties.values.filter { it.path.parentFile.toPath().equals(path.parent) };
+                if (parent.size > 1) {
+                    logger.debug { "Found ${parent.size} parents to attach test '${tp.name}' to, discarding binaries..." }
+                    parent = parent.filter { it.type != TargetProperties.Type.BIN };
+                }
                 if (parent.size != 1) {
-                    throw RuntimeException("Found '${parent.size}' parents to attach test '${tp.name}' to.")
+                    throw RuntimeException("Found ${parent.size} parents to attach test '${tp.name}' to: ${parent.map { it.name }}.")
                 }
```
dmitrii-ubskii, line 231: Ideally we'd sort out packages first so we don't have to do this, but that's a _much_ bigger change.

MODULE.bazel
```diff
@@ -62,10 +62,17 @@ register_toolchains("@llvm_toolchain//:all", dev_dependency = True)
 
 # Rust
 bazel_dep(name = "rules_rust", version = "0.56.0")
+git_override(
+    module_name = "rules_rust",
+    remote = "https://github.com/bazelbuild/rules_rust",
+    tag = "0.56.0",
+    patches = ["//patches:rules_rust.patch"],
+)
+
```
dmitrii-ubskii, line 72: This patch is backporting https://github.com/bazelbuild/rules_rust/pull/3251 from rules_rust 0.58. Unfortunately I couldn't simply upgrade rules_rust, because from 0.57 onwards it seems unable to unify crate dependencies between repos, which seems to lead to compile errors.

Another option is to upgrade to latest and apply this patch that fixes dependency unification: https://github.com/envoyproxy/toolshed/pull/3653, but this was the simpler change.

krishnangovindraj (approved):
lgtm. if it works, it works

MODULE.bazel
```diff
@@ -77,6 +84,9 @@ rust.toolchain(
 use_repo(rust, "rust_toolchains")
 register_toolchains("@rust_toolchains//:all")
 
+rust_host_tools = use_extension("@rules_rust//rust:extensions.bzl", "rust_host_tools")
+rust_host_tools.host_tools(version = "1.93.0")
```
dmitrii-ubskii, line 89: Needed for rules_rust to use `cargo` that understands edition2024

PR #604, opened 2026-04-22
Unpin and update transitive Rust dependencies

dmitrii-ubskii:
## Usage and product changes

We unpin the rust crate dependencies that were previously pinned to maintain compatibility with Rust 1.81. We also upgrade packages where possible without breaking changes.


farost:
@dmitrii-ubskii I remember you mentioned that pinning dependencies could be a better approach to avoid constant breaking changes after automatic cargo updates (and, while it's not a frequent issue, it would provide better security guarantees). Don't we want to consider this?

farost (approved):

dmitrii-ubskii:
> @dmitrii-ubskii I remember you mentioned that pinning dependencies could be a better approach to avoid constant breaking changes after automatic cargo updates (and, while it's not a frequent issue, it would provide better security guarantees). Don't we want to consider this?

It could, with two caveats:

- We can only pin dependencies that are _only_ used by the binary targets. Published libraries will likely create conflicts when pulled in larger projects. Our binary targets are the C driver (dylib counts), core (iffy, when embedded it would be a library too), and cluster.
- To my knowledge, none of our pinned dependencies were pinned for security reasons, but simply because we couldn't upgrade cargo to fix build issues with automatic upgrades.

PR #605, opened 2026-04-28
Add typedb-cluster dependencies

farost:
### Add dependencies needed for `typedb-cluster` 

Specifically:
* `raft-rs` -- library with an implementation of the Raft protocol
* `crc32fast` -- used for storage checksums
* `erased-serde` -- network response serialization
* `protobuf` -- `raft-rs` requirement
* `slog*` -- `raft-rs` logging requirement
* `tokio-test` / `tokio-util`-- new concurrency-related features and tests in `typedb`

### Loosen java_test requirements
Make `server_artifacts` optional for the tests, since they are not used in 3.x TypeDB driver tests.

### Fix raft proto builds in bazel                                                                                       
Bazel used to break three protobuf-related build patterns:
                                                                                                                           
1. `protobuf-build` sandbox incompatibility: `include_google_protos()` uses `std::env!("CARGO_MANIFEST_DIR")` -- a compile-time macro that bakes the sandbox path where protobuf-build was compiled. When a dependent crate's build script (e.g. `raft-proto`) runs in a different sandbox, that path no longer exists, and protoc cannot find `rustproto.proto`.   
2. `protobuf-build` modern-protoc rejection: `check_protoc_version()` requires `major == 3 && minor >= 1`, so any protoc from the 21+/29+ era (where protobuf bumped its major versioning) is rejected -- including the protoc Bazel itself builds (29.x).
3. `rust_tonic_compile` single-root limitation: the rule only passed the first proto source's `proto_source_root` to protoc, which breaks when proto files live in subdirectories with relative imports (e.g. import `raft_service/replication.proto` from `rpc/proto/`).

The issues from above were resolved.

## Implementation                                               
Add dependencies and run `update.sh`.

`protobuf-build` sandbox compatibility (`patches/protobuf-build-include.patch`):
- Patches `include_google_protos()` to check if the compile-time path exists first (preserving Cargo behavior), then falls back to a `PROTOBUF_BUILD_INCLUDE` env var resolved relative to the Bazel execroot                                 
- Generic: any crate using protobuf-build in Bazel can set this env var via `build_script_env``protobuf-build` modern-protoc compatibility (`patches/protobuf-build-protoc.patch`):                                    
- Loosens `check_protoc_version()` from `major == 3 && minor >= 1` to also accept `major >= 4`, so modern protoc (21.x,  29.x, ...) is accepted                                                                                                   
                                           
Crate annotations (`MODULE.bazel`):                                                                                      
- `protobuf-build`: applies the patches, exposes `include/**` as data
- `raft-proto`: brings `protobuf-build`'s include files into the build script sandbox via `build_script_data` and sets `PROTOBUF_BUILD_INCLUDE`; supplies protoc hermetically by adding `@@protobuf+//:protoc` to `build_script_data` and setting `PROTOC = "$(execpath @@protobuf+//:protoc)"`. No system protoc, no `PATH`, no `use_default_shell_env`. Required for macOS's stricter darwin-sandbox, which doesn't expose host PATH or the execroot to build scripts.                   
                                                                  
`rust_tonic_compile` configurable include roots (`builder/proto_grpc/rust/`):                                            
- Added `includes` attribute: explicit proto include paths relative to the workspace root
- When `includes` is provided, it is used instead of the auto-detected `proto_source_root` — this avoids duplicate symbol errors from protoc seeing the same file via multiple roots                                                              
- When `includes` is not provided, behavior is unchanged (auto-detect from `proto_source_root`)
- `compile.rs` updated to parse multiple semicolon-separated roots from `PROTOS_ROOT`                                    
- This allows proto files to use relative imports (e.g. import `raft_service/replication.proto`) that are compatible with both Cargo's `build.rs` and Bazel's `rust_tonic_compile`


farost:
Fixes typedb-cluster `bazel build //...`: https://github.com/typedb/typedb-cluster/commit/6f45d3bb132ce3c4f9e03cfc70fecde89bc35373 !

flyingsilverfin (approved):

dmitrii-ubskii (approved):
Seems straightforward

PR #609, opened 2026-05-23
Update maven dependency handling to only initialise the those used in the same repository

krishnangovindraj:
## Usage and product changes
Update maven dependency handling to only initialise the those used by java targets in this repository

## Motivation
Faster sync phase.

## Implementation
Apparently, the tools can resolve maven dependencies based on what's in the MODULE.bazel in this repository. So we reduce the set of dependencies to just those used by the java targets in this repository.

Adds a CI job that applies a patch and checks the full set of maven dependencies so we fail-fast in case we set an invalid version.


krishnangovindraj:
Subsumed in #611

PR #611, opened 2026-05-26
Speed up CI with various fixes

krishnangovindraj:
## Usage and product changes
Custom maven namespaces for smaller fetch, bump protobuf dependency for prebuilt `protoc`. Fixes for `rules_python` update

## Implementation
Groundwork for splitting out maven repos into various namespaces, avoiding computing the transitive dependencies of the 250+ maven targets we directly depend on somewhere in our codebase.


builder/proto_grpc/rust/compile.bzl
krishnangovindraj, line 1: This is Claude magic.
Bumping the protobuf gets the compiled `protoc` into the toolchain and this is supposedly the right way to use it.

library/ortools/deps.bzl
flyingsilverfin, line 1: These can be deleted in future

tool/maven_dependencies.MODULE.bazel
flyingsilverfin, line 1: This shouldn't be under tool, probably under `//dependencies`?

builder/swig/python.bzl
krishnangovindraj, line 1: More Claude magic to fix the error we got trying to support python3.14
Reasoning was that specifying this was linking it to a specific filename (`libpython3.11.so.1`) rather than whatever is correctly and automatically picked.

tool/maven_dependencies.MODULE.bazel
flyingsilverfin: Oh i see that we've put one in each subdirectory interesting

MODULE.bazel
```diff
@@ -157,8 +158,13 @@ use_repo(pip, "typedb_dependencies_ci_pip")
 bazel_dep(name = "rules_jvm_external", version = "6.6")
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_shell", version = "0.3.0")
+
 maven = use_extension("//library/maven:maven_extension.bzl", "maven")
-use_repo(maven, "maven")
+include("//tool:maven_dependencies.MODULE.bazel")
+include("//builder:maven_dependencies.MODULE.bazel")
+include("//factory:maven_dependencies.MODULE.bazel")
+include("//common:maven_dependencies.MODULE.bazel")
```
flyingsilverfin, line 166: Do you wanna do one local set of maven deps in //dependencies?

builder/maven_dependencies.MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+maven = use_extension("//library/maven:maven_extension.bzl", "maven")
+maven.namespace(name = "typedb_maven", artifacts = [
```
farost, line 6: How does it specifically help?

builder/maven_dependencies.MODULE.bazel
krishnangovindraj, line 1: New style I'm trying out with included `MODULE.bazel` so we don't lose track of what tool needs what dependency. We could actually use includes all over to make `MODULE.bazel` less biggg

MODULE.bazel
krishnangovindraj: Not really. I prefer this. Prevents us from having a massive monolithic list that we never remove things from even when we delete the tools.

.factory/automation.yml
```diff
@@ -10,6 +10,12 @@ build:
         sudo apt-get install -y libtinfo5
         bazel build //...
         bazel test $(bazel query 'kind(checkstyle_test, //... except tool/ide/...)') || exit 1
+    validate-all-maven-deps-build:
```
flyingsilverfin, line 13: can we roll into build job?

builder/maven_dependencies.MODULE.bazel
krishnangovindraj: The time to "fetch" the `@maven` repository goes down from a minute (on my mac) to seconds.

.factory/automation.yml
krishnangovindraj: No. The build tests this as it's consumed by downstream repos.
This job just makes sure ALL the declared versions are available.

flyingsilverfin (approved):

PR #615, opened 2026-06-08
Inline maven dependencies

krishnangovindraj:
## Usage and product changes
Inline maven dependency declarations as `include` can't be used outside ROOT modules.


farost (approved):
Sad that the `include`s don't work

PR #623, opened 2026-07-17
Make librocksdb-sys and lz4-sys C deterministic to improve caching

flyingsilverfin:
## Usage and product changes

Make `-sys` crates reproducible and deterministic, improving caching. This will apply on mac/linux, but not apply on Windows (builds will not break on windows).

## Motivation

Speed up caching across different caches and machines.

## Implementation

librocksdb-sys and lz4-sys compile vendored C through the bazel cc toolchain, which bakes the absolute sandbox exec-root path (.../processwrapper-sandbox/<N>/...) into every object file's debug info. The <N> counter changes each run, so the build-script outputs are never byte-identical across executions. That drift breaks the disk/remote cache key for librocksdb-sys and everything downstream of RocksDB whenever an upstream -sys crate re-executes.

Appending `-ffile-prefix-map=${pwd}=.` to the build-script CFLAGS/CXXFLAGS make the embedded paths and artifacts deterministic. rules_rust's _merge_env_dict appends these to the toolchain flags, preserving the __DATE__/__TIME__ redaction.

Note: MSVC treats -ffile-prefix-map as a non-fatal warning and keeps compiling.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -7,7 +7,7 @@ build:
     build:
       image: typedb-ubuntu-22.04
       command: |
-        sudo apt-get install -y libtinfo5
+        sudo apt-get update && sudo apt-get install -y libtinfo5
         if [ -n "${CARGO_BAZEL_GENERATOR_URL_LINUX_X86_64:-}" ] && [ -n "${CARGO_BAZEL_GENERATOR_SHA256_LINUX_X86_64:-}" ]; then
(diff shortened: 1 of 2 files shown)


.factory/automation.yml
```diff
@@ -7,7 +7,7 @@ build:
     build:
       image: typedb-ubuntu-22.04
       command: |
-        sudo apt-get install -y libtinfo5
+        sudo apt-get update && sudo apt-get install -y libtinfo5
```
flyingsilverfin, line 10: Required to make CI red - ubuntu had moved this dep into another place and needed update to update package lists

